PR #4, opened 2021-02-02
Configure an empty cluster

tomjelinek:
Implementing support for configuring an empty cluster, i.e. cluster running no resources / workload.


richm:
re: ansible-lint issue 301 - Commands should not change things if nothing needs doing

tasks that get information from the managed system should use `changed_when: false` to denote that these tasks never change the persistent state of the managed system - or in some cases, use `when: condition` in case the task only needs to be run in certain conditions

re: ansible-lint issue 503 - Tasks that run when changed should likely be handlers

Typically you will let ansible restart your daemons after the play is complete, using a handler - but if you need to restart your daemon before you can proceed to other tasks, but you only want to restart the daemon in case something has changed (i.e. you do not want to restart your daemon unnecessarily), then you'll have to manage that in the code, and add `'503'` to the skip list in your .ansible-lint - see https://github.com/linux-system-roles/metrics/commit/4836c132c8d9ba9e7a27ca5e77be402e980d3278 for an example of the metrics role needing to restart pmcd before proceeding to other tasks, but only doing the restart if something changed.

re: ansible-lint issue 208 File permissions unset or incorrect

You'll need to add `owner`, `group`, and `mode` for files and directories created by the role

README.md
```diff
@@ -3,74 +3,204 @@
...
+
+#### `ha_cluster_fence_agent_packages`
+
+list of fence agent packages to install, default: fence-agents-all, fence-virt
+
+#### `ha_cluster_hacluster_password`
+
+string, no default - must be specified
+
+Password of the `hacluster` user. This user has full access to a cluster.
```
richm, line 42: add text like "Recommend vault encrypting the value. See https://docs.ansible.com/ansible/latest/user_guide/vault.htm"

README.md
```diff
@@ -3,74 +3,204 @@
...
+Authentication and encryption key for Corosync communication. It is highly
+recommended to have a unique value for each cluster. The key should be 256
+bytes of random data.
+
+If no key is specified, a key already present on the nodes will be used. If
+nodes don't have the same key, a key from one node will be distributed to other
+nodes so that all nodes have the same key. If no node has a key, a new key will
+be generated and distributed to the nodes.
+
+If this variable is set, `ha_cluster_regenerate_keys` is ignored for this key.
```
richm, line 57: add text like "If value is provided, recommend vault encrypting the value. See https://docs.ansible.com/ansible/latest/user_guide/vault.htm"

README.md
```diff
@@ -3,74 +3,204 @@
...
+Authentication and encryption key for Pacemaker communication. It is highly
+recommended to have a unique value for each cluster. The key should be 256
+bytes of random data.
+
+If no key is specified, a key already present on the nodes will be used. If
+nodes don't have the same key, a key from one node will be distributed to other
+nodes so that all nodes have the same key. If no node has a key, a new key will
+be generated and distributed to the nodes.
+
+If this variable is set, `ha_cluster_regenerate_keys` is ignored for this key.
```
richm, line 72: add text like "If value is provided, recommend vault encrypting the value. See https://docs.ansible.com/ansible/latest/user_guide/vault.htm"

README.md
```diff
@@ -3,74 +3,204 @@
...
+path to fence-virt or fence-xvm pre-shared key file, default: `null`
+
+Authentication key for fence-virt or fence-xvm fence agent.
+
+If no key is specified, a key already present on the nodes will be used. If
+nodes don't have the same key, a key from one node will be distributed to other
+nodes so that all nodes have the same key. If no node has a key, a new key will
+be generated and distributed to the nodes.
+
+If this variable is set, `ha_cluster_regenerate_keys` is ignored for this key.
```
richm, line 85: add text like "If value is provided, recommend vault encrypting the value. See https://docs.ansible.com/ansible/latest/user_guide/vault.htm"

README.md
```diff
@@ -3,74 +3,204 @@
...
+#### `ha_cluster_pcsd_public_key_src`, `ha_cluster_pcsd_private_key_src`
+
+path to pcsd TLS certificate and key, default: `null`
+
+TLS certificate and private key for pcsd. If this is not specified, a
+certificate - key pair already present on the nodes will be used. If
+certificate - key pair is not present, a random new one will be generated.
+
+If these variables are set, `ha_cluster_regenerate_keys` is ignored for this
+certificate - key pair.
```
richm, line 99: add text like "If private key value is provided, recommend vault encrypting the value. See https://docs.ansible.com/ansible/latest/user_guide/vault.htm"

tasks/cluster-enable-disable.yml
```diff
@@ -0,0 +1,26 @@
...
+    # boot. Null (nochange) could lead to situations where part of cluster
+    # nodes are configured to start the cluster on boot while other nodes in
+    # the same cluster are not.
+    - item != "corosync-qdevice" or __ha_cluster_qdevice_in_use
+
+- name: Get services status - detect SBD
+  service_facts:
+
+# The role does not support configuring SBD yet, therefore we always disable it.
+- name: Enable or disable SBD
```
richm, line 21: disable SBD

tasks/cluster-enable-disable.yml
```diff
@@ -0,0 +1,26 @@
+- name: Enable or disable configured cluster services on boot
+  service:
+    name: "{{ item }}"
+    enabled: "{{ ha_cluster_start_on_boot }}"
+  loop:
+    - corosync
+    - corosync-qdevice
+    - pacemaker
```
richm, line 8: These should be defined as a list `__ha_cluster_services` in `vars/main.yml`

tasks/cluster-start-and-reload.yml
```diff
@@ -0,0 +1,114 @@
...
+    state: stopped
+  loop:
+    - pacemaker
+    - corosync-qdevice
+    - corosync
+  when:
+    - >
+        __ha_cluster_distribute_corosync_conf.changed
+        or __ha_cluster_distribute_corosync_authkey.changed
+        or __ha_cluster_distribute_pacemaker_authkey.changed
```
richm, line 34: ok - then you can add `'503'` to .ansible-lint to suppress the error

tasks/cluster-enable-disable.yml
richm: hm - on the other hand, I see several places below where you want to treat each service individually - I think it may still be worth it to define all of the services in a single list in cases where you need to operate on all of them in a single task

tasks/cluster-start-and-reload.yml
```diff
@@ -0,0 +1,114 @@
...
+
+- name: Start pacemaker
+  service:
+    name: pacemaker
+    state: started
+
+
+- name: Wait for the cluster to fully start and form membership
+  command:
+    cmd: pcs cluster start --all --wait
```
richm, line 71: you don't need a separate `cmd` element, just `command: pcs cluster start --all --wait`

do you need a timeout for this task e.g. if this task hasn't completed after 15 minutes, then report failure?

tasks/cluster-start-and-reload.yml
```diff
@@ -0,0 +1,114 @@
...
+  changed_when: no
+
+
+# Purge removed nodes from Pacemaker's caches
+
+- name: List pacemaker nodes
+  shell:
+    cmd: >
+      crm_mon -X
+      | xmllint --xpath '/crm_mon/nodes/node/@name' -
```
richm, line 87: is `xmllint` a package dependency of the other pacemaker packages?  If not, you'll need to add the package to the list of packages that this role manages.

tests/files/corosync-authkey
```diff
@@ -0,0 +1,3 @@
+,��`�W���?��h�y+,��-U"�B0����2��!����[>�a!<ɹZ/2���e��Fh�o5П�w�O��iHj�\=��GK	[O�/
+y���7S#r��Zטŷ�:�ٹ���R�L.�" �	��ޯk�o����#]��nҸ��Tݘ
+�O��Q�Iǽo��~<-��PI9��րA��.w�<�BfK�/�Yc����4*�i��("m���6��u@sQ����	��m;?�
```
richm, line 3: do you need to add binary files for testing?  It would be better if the test generated these rather than having hardcoded keys checked into git

tests/files/pcsd.crt
```diff
@@ -0,0 +1,26 @@
...
+BgkqhkiG9w0BAQsFAAOCAYEAh7zwVTLNO8kgE9Nj3cbXAJMEJ6ZoWp3Up+Q19F7x
+Vg3cJt3va/BWmk01ihjr6kXhpPiSZl/+FyQocU5VghxjhTlobzsgKQhsBR0WkBWX
+b6u296J2wybbukuecR000TBUGooHaNetu9KmTS5nW5e7i6tJeztZEg6MndRx1vVD
+bTrKQDqH9eSGj1UcuhT9okRBl6oXavFb8/buweCUptpGfozA8XqX++2UZPo2hR0v
+t1aI2y2LOAb3Iwt997yAI9GIY11uPGJu4sbe4yhpTFM2xh5t3ytPJHTNGk3/poO/
+O9Qt3Dgrp9gCJhZ38gzP+NQwbBOZq6PJXV+4MsRMWuAjYNpiXywmT5z6C362WzEs
+A9/p66zYzpu614GPvQVhkyZ+kpCDD4rqQe3rAu7bWOwfG6q9HFmhpMDloWipLhp0
+p7piYlrotq/Um+Y8tcXxFDTS6Kmi2a3r9g08VP5+qOw0hxwTEImsqSFiy5pGFm7S
+0Lrr3QHpCw+kj3hD8Tap0/iN
+-----END CERTIFICATE-----
```
richm, line 26: https://github.com/linux-system-roles/logging/blob/master/tests/tasks/create_tests_certs.yml is an example of how a test can generate certs/keys

tests/tests_cluster_basic.yml
```diff
@@ -0,0 +1,73 @@
+- name: Minimal cluster configuration
+  hosts: all
+  vars_files: vars.yml
```
richm, line 3: please move this into a `tests/vars/` subdirectory

tests/tests_cluster_basic.yml
```diff
@@ -0,0 +1,73 @@
...
+            - /etc/pacemaker/authkey
+            - /etc/cluster/fence_xvm.key
+            - /var/lib/pcsd/pcsd.key
+            - /var/lib/pcsd/pcsd.crt
+
+        - name: Run HA Cluster role
+          include_role:
+            name: linux-system-roles.ha_cluster
+
+        - meta: flush_handlers
```
richm, line 24: I don't think you need this since you are not using handlers

README.md
tomjelinek: fixed in 1c2a9bf

README.md
tomjelinek: fixed in 1c2a9bf

README.md
tomjelinek: fixed in 1c2a9bf

README.md
tomjelinek: fixed in 1c2a9bf

README.md
tomjelinek: fixed in 1c2a9bf

tasks/cluster-enable-disable.yml
tomjelinek: fixed in 1c2a9bf

tasks/cluster-enable-disable.yml
tomjelinek: fixed in 1c2a9bf

tests/tests_cluster_basic.yml
tomjelinek: fixed in 1c2a9bf

tests/tests_cluster_basic.yml
tomjelinek: fixed in 1c2a9bf

tests/files/corosync-authkey
tomjelinek: You are absolutely right, fixed in 6ac095a

tests/files/pcsd.crt
tomjelinek: Thanks for the example, fixed in 6ac095a

tasks/cluster-start-and-reload.yml
tomjelinek: fixed in 53ff3e8

tasks/cluster-start-and-reload.yml
tomjelinek: Timeout is built in the pcs command and it exits with nonzero if the cluster is not started when the timeout passes. I believe that should make ansible report failure.

tasks/cluster-start-and-reload.yml
tomjelinek: `xmllint` command is provided by `libxml2` package which is a dependency of several HA Cluster packages

tomjelinek:
tox is all green now. I'm moving to centos-8 / rhel-8 / fedora issues.

richm:
```
TASK [linux-system-roles.ha_cluster : Install cluster packages] ****************
task path: /tmp/tmp5_g_826r/tasks/install-and-configure-packages.yml:1
fatal: [/cache/centos-8.qcow2]: FAILED! => {"changed": false, "failures": ["No package corosync available.", "No package libknet1-plugins-all available.", "No package pacemaker available.", "No package pcs available."], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}
```
do we need to enable additional repos on centos8 and other platforms?

tomjelinek:
Most of the HA packages are provided in HA Addon repos. I've discussed that with @pcahyna already, the resolution is that it's the responsibility of the role to enable those repos. Do you agree with that, @richm?

tasks/cluster-setup-pcs-0.10.yml
```diff
@@ -0,0 +1,22 @@
+- name: Create a corosync.conf file content using pcs-0.10
+  command:
+    cmd: >
+      pcs cluster setup
+      --corosync_conf {{ __ha_cluster_tempfile_corosync_conf.path | quote }}
+      --overwrite
+      {{ ha_cluster_cluster_name | quote }}
+      {% for node in ansible_play_hosts %}
```
richm, line 8: I don't think this is a task that we can test in our CI - in CI, the ansible hostname is the path and filename of the platform.qcow2 e.g. `/cache/centos8.qcow2`

note that the vpn role has a similar problem - it needs to test setting up configuration of tunnels with actual hostnames.  This was solved by having the test change the inventory_hostname, and using add_host to create some dummy hosts.
see https://github.com/linux-system-roles/vpn/blob/master/tests/tasks/setup_test.yml and https://github.com/linux-system-roles/vpn/blob/master/tests/tasks/add_hosts.yml - they are called from the test like this: https://github.com/linux-system-roles/vpn/blob/master/tests/tests_host_to_host_cert.yml#L12
however - I'm not sure if that changes `ansible_play_hosts` - in your testing you might have to overwrite `ansible_play_hosts` with your "fake" hostnames.

richm:
> Most of the HA packages are provided in HA Addon repos. I've discussed that with @pcahyna already, the resolution is that it's the responsibility of the role to enable those repos. Do you agree with that, @richm?

Yes

tasks/cluster-setup-pcs-0.10.yml
tomjelinek: Hopefully, I managed to resolve this - fedora-32/ansible-2.9 and fedora-33/ansible-2.9 are passing now.

tasks/check-and-prepare-role-variables.yml
```diff
@@ -9,20 +9,23 @@
...
+        hostvars[inventory_hostname].ha_cluster.node_name \
+        | default(inventory_hostname) \
       }}"
-  loop: "{{ ansible_play_hosts }}"
-  run_once: yes
+
+- name: Extract cluster node names
+  set_fact:
+    __ha_cluster_all_node_names: "{{ hostvars[item].__ha_cluster_node_name }}"
+  loop: "{{ ansible_play_hosts_all }}"
```
richm, line 22: I'm not sure what you are trying to do here but it doesn't look right to me - `__ha_cluster_all_node_names` will not be a list, it will be a string - can you explain what you are trying to do here?

tomjelinek: There are three tasks which work together to construct the list of node names.

`Discover cluster node names` - figures out a node name for each host, that is either an inventory variable or `inventory_hostname`. This is actually used in tests to replace `/cache/centos8.qcow2` with node names sensible from the HA cluster point of view. It also ensures that `__ha_cluster_node_name` is always defined.
```
ok: [rh83-node2] => {"ansible_facts": {"__ha_cluster_node_name": "rh83-node2"}, "changed": false}
ok: [rh83-node3] => {"ansible_facts": {"__ha_cluster_node_name": "rh83-node3"}, "changed": false}
```

`Extract cluster node names` - by looping over `ansible_play_hosts_all`, I get all node names into one 'variable'. The fact itself is not important here at all and it gets overwritten by the following task. What matters is the registered variable `__ha_cluster_node_name_result`, which is processed by the following task. Yes, `__ha_cluster_all_node_names` will be a string, but it will be a string for each node.
```
ok: [rh83-node2] => (item=rh83-node2) => {"ansible_facts": {"__ha_cluster_all_node_names": "rh83-node2"}, "ansible_loop_var": "item", "changed": false, "item": "rh83-node2"}
ok: [rh83-node2] => (item=rh83-node3) => {"ansible_facts": {"__ha_cluster_all_node_names": "rh83-node3"}, "ansible_loop_var": "item", "changed": false, "item": "rh83-node3"}
ok: [rh83-node3] => (item=rh83-node2) => {"ansible_facts": {"__ha_cluster_all_node_names": "rh83-node2"}, "ansible_loop_var": "item", "changed": false, "item": "rh83-node2"}
ok: [rh83-node3] => (item=rh83-node3) => {"ansible_facts": {"__ha_cluster_all_node_names": "rh83-node3"}, "ansible_loop_var": "item", "changed": false, "item": "rh83-node3"}
```
`Collect cluster node names` - collects node names from the registered result of `Extract cluster node names` to create a list of strings - node names. Since this works based on `__ha_cluster_node_name`, which is always defined by the first task, `__ha_cluster_all_node_names` is always defined as well and it contains all nodes.
```
ok: [rh83-node2] => {"ansible_facts": {"__ha_cluster_all_node_names": ["rh83-node2", "rh83-node3"]}, "changed": false}
ok: [rh83-node3] => {"ansible_facts": {"__ha_cluster_all_node_names": ["rh83-node2", "rh83-node3"]}, "changed": false}
```

Perhaps I could have named the fact in `Extract cluster node names` differently to make it more clear.

It seems quite complicated to me to have these three tasks to create a single list of node names. It's the best I was able to figure out, but my ansible skills are still not the best. Let me know if you know a simpler way to achieve this.

richm: ok - so you need to have some variable e.g. `__ha_cluster_all_node_names` in the hostvars of each host.  This variable contains the list of all of the ha_cluster node names for all of the hosts in the cluster.

I believe `hostvars[inventory_hostname].ha_cluster.node_name` is exactly the same as `ha_cluster.node_name`.

Let me see what I can come up with.

richm: you only need 1 task
```yaml
- name: Collect cluster node names
  set_fact:
    __ha_cluster_all_node_names: "{{
      ansible_play_hosts_all | map('extract', hostvars, ['ha_cluster', 'node_name']) |
      map('default', inventory_hostname) | list
    }}"
```
See https://docs.ansible.com/ansible/2.8/user_guide/playbooks_filters.html#extracting-values-from-containers
Here is some sample code https://gist.github.com/richm/0f6dd759a461705945eb5dd0c5287815
I run it like this: `ANSIBLE_STDOUT_CALLBACK=debug ansible-playbook -vv ha_cluster_test.yml`
the output looks like this:
```
TASK [debug] *******************************************************************
task path: /home/rmeggins/ansible_sandbox/ha_cluster_test.yml:19
ok: [localhost] => {}

MSG:

ansible_play_hosts_all ['localhost', 'node1', 'node3', 'node2'] hostvars ['localhost', 'node1-ha', 'node3-ha', 'localhost']
```
I'm assuming in a "real" environment there won't be a `localhost` nor duplicate hosts - if duplication is a problem, there's probably something else going on, but at any rate you can use the `unique` filter if you think it could be a problem

pcahyna:
@jharuda some problem with the staging test-harness - fails very early:
```
linux-system-roles-staging-117-pf8rl: linux-system-roles/ha_cluster: pull #4 (fa339dc) on fedora-32
===================================================================================================

Traceback (most recent call last):
  File "/test/run-tests", line 1035, in handle_task
    result = task.run(
  File "/test/run-tests", line 476, in run
    image_path = fetch_image(image_url, cachedir, self.image["name"])
  File "/test/run-tests", line 297, in fetch_image
    image_last_modified_by_src = get_metadata_from_url(url, "Last-Modified")
  File "/test/run-tests", line 276, in get_metadata_from_url
    return dict(url_response.getheaders())[metadata_key]
KeyError: 'Last-Modified'
```

jharuda:
[citest bad]

richm:
> @jharuda some problem with the staging test-harness - fails very early:
> 
> ```
> linux-system-roles-staging-117-pf8rl: linux-system-roles/ha_cluster: pull #4 (fa339dc) on fedora-32
> ===================================================================================================
> 
> Traceback (most recent call last):
>   File "/test/run-tests", line 1035, in handle_task
>     result = task.run(
>   File "/test/run-tests", line 476, in run
>     image_path = fetch_image(image_url, cachedir, self.image["name"])
>   File "/test/run-tests", line 297, in fetch_image
>     image_last_modified_by_src = get_metadata_from_url(url, "Last-Modified")
>   File "/test/run-tests", line 276, in get_metadata_from_url
>     return dict(url_response.getheaders())[metadata_key]
> KeyError: 'Last-Modified'
> ```

fixed in https://github.com/linux-system-roles/test-harness/pull/140 - I guess isn't rolled out everywhere yet

tasks/check-and-prepare-role-variables.yml
tomjelinek: Thank you very much, this works perfectly! `Discover cluster node names` task is actually needed as well, because `__ha_cluster_node_name` fact it sets is used in the role. But we got rid of the ugly `Extract cluster node names` task.

`unique` is not used on purpose. If there is a duplicity in node names, it will get discovered and reported by `pcs` as an error. That is a correct behavior, node names in a cluster must be unique. If they are not, then there is either a bug in the role or the user provided duplicate names. Both cases should be reported as an error.

tomjelinek:
[citest bad]

tasks/check-and-prepare-role-variables.yml
```diff
@@ -0,0 +1,25 @@
...
+      when: lookup("vars", item, default="") | length < 1
+      loop:
+        - ha_cluster_hacluster_password
+      run_once: yes
+  when: ha_cluster_cluster_present
+
+- name: Discover cluster node names
+  set_fact:
+    __ha_cluster_node_name: "{{
+        hostvars[inventory_hostname].ha_cluster.node_name
```
richm, line 15: isn't `hostvars[inventory_hostname].ha_cluster.node_name` the exact same thing as `ha_cluster.node_name`?

tasks/install-and-configure-packages.yml
```diff
@@ -0,0 +1,27 @@
+- name: Enable standard repositories
+  rhsm_repository:
+    name: "{{ item }}"
+    state: enabled
+  loop: "{{ __ha_cluster_repos }}"
+  when:
+    - ha_cluster_enable_repos
+  ignore_errors: yes
```
richm, line 8: why do you need to use `ignore_errors: yes` here?

tomjelinek: It's needed to make the role work in CI systems which don't have access to RHEL subscriptions. This has been added based on a meeting with @pcahyna.

tasks/check-and-prepare-role-variables.yml
tomjelinek: Oh yes, of course it is. I should have realized that, sorry.

tasks/install-and-configure-packages.yml
richm: If we need some sort of hack, I would prefer to add some sort of extra variable like `ha_cluster_skip_ha_repos` that the CI system could add to tell this task to be skipped.
Or, make it so that the `rhsm_repository` task will see that the CI system has already enabled the necessary repos, so that it is a no-op and will not return an error.
Using `ignore_errors: true` in the actual production code is not good, unless there is just no other solution.

jharuda:
[citest bad]

tasks/install-and-configure-packages.yml
tomjelinek: There is such a variable: `ha_cluster_enable_repos: no`. I was not aware of the possibility of setting extra vars in CI. If you can do it, then that's a clean solution.

The long term solution, as discussed on the meeting, would be to make this actually work in CI, so the role works without workarounds.

richm: > There is such a variable: `ha_cluster_enable_repos: no`. I was not aware of the possibility of setting extra vars in CI. If you can do it, then that's a clean solution.

Well, you can do it, by adding `ha_cluster_enable_repos: no` to `tests/tests_*.yml`

> 
> The long term solution, as discussed on the meeting, would be to make this actually work in CI, so the role works without workarounds.

tasks/install-and-configure-packages.yml
```diff
@@ -18,8 +16,6 @@
   when:
     - ansible_distribution == "CentOS"
     - ha_cluster_enable_repos
```
richm, line 18: yes
then add `ha_cluster_enable_repos: false` to `tests/tests_*.yml`

jharuda:
[citest bad]

tasks/install-and-configure-packages.yml
tomjelinek: > Well, you can do it, by adding ha_cluster_enable_repos: no to tests/tests_*.yml

If I do it like that, then anyone running the tests would have to enable the repos manually. Is this what we want? My impression was we were trying to work around the issue just in this CI here and let the role enable repos everywhere else. That's why it haven't occur to me to do it like that. But if you think this is the correct way to go, I'm fine with that.

richm: > > Well, you can do it, by adding ha_cluster_enable_repos: no to tests/tests_*.yml
> 
> If I do it like that, then anyone running the tests would have to enable the repos manually. Is this what we want?

For now, yes.

> My impression was we were trying to work around the issue just in this CI here and let the role enable repos everywhere else. That's why it haven't occur to me to do it like that. But if you think this is the correct way to go, I'm fine with that.

If we find that this is not suitable, we can change it later.

There's also another problem with `rhsm_repository` - in Ansible 2.10 and later, it looks like it has moved out of core and into community support - https://docs.ansible.com/ansible/latest/collections/community/general/index.html#plugins-in-community-general - which means a dependency on the community.general collection

Furthermore - this so far is the *only* role that has an explicit dependency on an external vendor packaging system - I believe in the other roles, we just mention in the README that there is xxx additional dependency/repo that has to be enabled - so I'm not quite sure how to handle this case.

richm:
[citest bad]

tasks/install-and-configure-packages.yml
tomjelinek: Ok, I've updated the tests so they set `ha_cluster_enable_repos`.

There is an idea to create a separate role for enabling repositories which other roles could use. SAP role also enables repositories and there may be even other roles later which could benefit from that. Personally, I support this idea as that would concentrate all these repositories-related tasks to one place. In the meantime, we can either stick with having `rhsm_repository` in `ha_cluster` role or drop it and note the need for enabling the repo in the README. The readme way is easier for me but I don't feel to be in a position to decide this by myself.

tomjelinek:
[citest bad]

tomjelinek:
Several tests have failed due to OOM killer killing processes on the machines. Would it be possible to increase the amount of memory for the machines?

tomjelinek:
rhel-8 and rhel-x tests are failing, because they don't have access to HA cluster packages. I'm not sure why that is, as rhel-x and rhel-8 staging tests are passing.

pcahyna:
> rhel-8 and rhel-x tests are failing, because they don't have access to HA cluster packages. I'm not sure why that is, as rhel-x and rhel-8 staging tests are passing.

I think it is because @jharuda has not deployed the same change to enable HA repos in production. Jakub, when you deploy production, please rerun the failed tests.

richm:
> Several tests have failed due to OOM killer killing processes on the machines. Would it be possible to increase the amount of memory for the machines?

I would like to investigate these further.  Which tests are these?

pcahyna:
> Several tests have failed due to OOM killer killing processes on the machines. Would it be possible to increase the amount of memory for the machines?

Can be done using provision.fmf

richm:
[citest]

tomjelinek:
>> Several tests have failed due to OOM killer killing processes on the machines. Would it be possible to increase the amount of memory for the machines?

> I would like to investigate these further. Which tests are these?

Unfortunately, the tests have been restarted in the meantime. These are the only artifacts with OOM killer I can see now:
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-4-18066ad-centos-8-20210216-121111/artifacts/

richm:
> > > Several tests have failed due to OOM killer killing processes on the machines. Would it be possible to increase the amount of memory for the machines?
> 
> > I would like to investigate these further. Which tests are these?
> 
> Unfortunately, the tests have been restarted in the meantime. These are the only artifacts with OOM killer I can see now:
> https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-4-18066ad-centos-8-20210216-121111/artifacts/

ok - yes, looks like pacemaker needs more ram - take a look at certificate role - https://github.com/linux-system-roles/certificate/tree/master/tests - create a tests/provision.fmf file like that for the ha_cluster role

tests/provision.fmf
```diff
@@ -0,0 +1,3 @@
+standard-inventory-qcow2:
+  qemu:
+    m: 2048
```
richm, line 3: This is the default https://pagure.io/standard-test-roles/blob/master/f/inventory/standard-inventory-qcow2#_336

tasks/install-and-configure-packages.yml
richm: Ok.  Let's leave the code as-is with rhsm_repository.  We can revisit this later once we get some feedback from customers, or when we have rhn/rhsm roles.

tests/provision.fmf
tomjelinek: It seems to me this machine, where OOM killer got runned, had only 1GB of RAM:
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-4-18066ad-centos-8-20210216-121111/artifacts/centos-8.qcow2.guest.log
That's why I set it to 2GB.

richm: > That's why I set it to 2GB.

Hmm - that's disturbing that we cannot rely on the default set in standard-inventory-qcow2!  Sounds like a bug in that script.

So ok - for ha_cluster - let's keep this provision.fmf just to make sure

pcahyna:
> > > Several tests have failed due to OOM killer killing processes on the machines. Would it be possible to increase the amount of memory for the machines?
> 
> > I would like to investigate these further. Which tests are these?
> 
> Unfortunately, the tests have been restarted in the meantime. These are the only artifacts with OOM killer I can see now:
> https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-4-18066ad-centos-8-20210216-121111/artifacts/

Looking closer, that test is for the centos7 controller. The inventory script in standard-test-roles in EPEL7 is old and the default there is 1 GB, which explains it. We need to update the inventory script :-(. For now, keep provision.fmf as workaround.

tests/provision.fmf
pcahyna: @richm see my last comment

richm:
lgtm

PR #11, opened 2021-03-30
Fix issues related to enabling repositories

tomjelinek:
This pull request fixes two issues:
* In CentOS 8.3, the `HighAvailability` repository got renamed to `ha`. I updated the role to reflect that.
* Tasks for enabling repositories have been extracted to separate files. This way, tasks for new platform/version can be added in the same way as variable files.

### DIFF
--- a/tasks/enable-repositories/CentOS.yml
+++ b/tasks/enable-repositories/CentOS.yml
@@ -1,14 +1,28 @@
 # SPDX-License-Identifier: MIT
 ---
+- name: List active CentOS repositories
+  command:
+    cmd: dnf repolist
+    # See the comment bellow.
+    warn: no
+  register: __ha_cluster_repolist
+  changed_when: no
+  check_mode: no
+
 - name: Enable CentOS repositories
   command:
-    cmd: "dnf config-manager --set-enabled {{ item | quote }}"
-    # There doesn't seem to be a plugin for enabling already defined repos. So
-    # we silence the warning which recommends using an ansible plugin instead
-    # of running dnf command.
+    cmd: "dnf config-manager --set-enabled {{ item.id | quote }}"
+    # Silence a warning which recommends using ansible dnf module instead of
+    # running dnf command.
+    #
+    # yum_repository module documentation states, "If you wish to update an
+    # existing repository definition use ini_file instead."
+    #
+    # ini_file requires a file path and a section name, both of which are
+    # different in various CentOS minor versions.
+    #
+    # Using dnf, similar differences must also be handled. But at least the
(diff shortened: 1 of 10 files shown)


tasks/enable-repositories/CentOS.yml
```diff
@@ -0,0 +1,14 @@
...
+    cmd: "dnf config-manager --set-enabled {{ item | quote }}"
+    # There doesn't seem to be a plugin for enabling already defined repos. So
+    # we silence the warning which recommends using an ansible plugin instead
+    # of running dnf command.
+    warn: no
+  loop: "{{ __ha_cluster_repos }}"
+  # The dnf command does not indicate if any changes have been made. To silence
+  # ansible lint (301 Commands should not change things if nothing needs doing),
+  # changed_when is set explicitly.
+  changed_when: yes
```
richm, line 14: > There doesn't seem to be a plugin for enabling already defined repos

Have you tried https://docs.ansible.com/ansible/2.8/modules/yum_repository_module.html#yum-repository-module ?

The use of `changed_when` breaks idempotency - the role will never be idempotent.  In order to be idempotent, if you cannot use `yum_repository` you'll have to use `command` or `shell` to check if the repos are enabled first, then only execute the `    cmd: "dnf config-manager --set-enabled {{ item | quote }}"` if the repo isn't enabled.

tomjelinek: Yes, I tried `yum repository`. It is not capable of changing an already existing repository.  For it to work, I had to write a complete repository definition. That approach seemed fragile to me as the HA repo file is different in various CentOS minor versions.

I added tasks to check if repositories are already enabled.

richm (approved):
lgtm

richm:
@tomjelinek I fixed the tox test issues - please rebase and repush your PR

richm:
[citest bad]

tasks/enable-repositories/Fedora.yml
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: MIT
+---
+# No enable-repositories tasks are needed for Fedora. This file servers as a
```
richm, line 3: `servers` -> `serves`

richm:
fedora 32 fails - is this a known problem?

tasks/enable-repositories/Fedora.yml
tomjelinek: Thanks, fixed.

tomjelinek:
> fedora 32 fails - is this a known problem?

There seem to be dependency issues related to fence-agents packages. Those packages are out of my scope. I asked fence-agents maintainer to take a look at it and let us know what may be the problem.

PR #14, opened 2021-04-09
fix reading preshared keys

tomjelinek:
lookup('file') does not work with binary files and causes failures depending on the content of a preshared key file

### DIFF
--- a/tasks/presharedkey.yml
+++ b/tasks/presharedkey.yml
@@ -5,13 +5,32 @@
 # key from cluster nodes. If no node has a key, a random key generated in step
 # 2 is going to be used.
 
-- name: "Fetch {{ preshared_key_label }} if provided"
-  set_fact:
-    __ha_cluster_some_preshared_key: >
-      "{{ lookup('file', preshared_key_src) | b64encode }}"
+- name: "Retrieve {{ preshared_key_label }} from the controller"
+  block:
+    - name: "Check if {{ preshared_key_label }} exists on the controller"
+      stat:
+        path: "{{ preshared_key_src }}"
+      register: __ha_cluster_preshared_key_stat_controller
+
+    - name: "Slurp existing {{ preshared_key_label }} from the controller"
+      slurp:
+        src: "{{ preshared_key_src }}"
+      register: __ha_cluster_preshared_key_slurp_controller
+      when: __ha_cluster_preshared_key_stat_controller.stat.exists
+      # Prevent key contents to be printed to the output
+      no_log: true
+
+    - name: "Use the slurped {{ preshared_key_label }} from the controller"
+      set_fact:
+        __ha_cluster_some_preshared_key: >
+          "{{ __ha_cluster_preshared_key_slurp_controller.content }}"
+      when: __ha_cluster_preshared_key_stat_controller.stat.exists
+      # Prevent key contents to be printed to the output
+      no_log: true
   when:
     - preshared_key_src is string and preshared_key_src | length > 1
   run_once: yes
+  delegate_to: localhost
 
 - name: "Generate random {{ preshared_key_label }}"
   set_fact:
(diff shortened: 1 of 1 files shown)


tomjelinek:
centos-8 is expected to fail, a fix for that is in #11

tasks/presharedkey.yml
```diff
@@ -5,13 +5,28 @@
...
+    - name: "Check if {{ preshared_key_label }} exists on the controller"
+      stat:
+        path: "{{ preshared_key_src }}"
+      register: __ha_cluster_preshared_key_stat_controller
+
+    - name: "Slurp existing {{ preshared_key_label }} from the controller"
+      slurp:
+        src: "{{ preshared_key_src }}"
+      register: __ha_cluster_preshared_key_slurp_controller
+      when: __ha_cluster_preshared_key_stat_controller.stat.exists
```
richm, line 19: Please use `no_log: true` so that the key contents won't be accidentally printed to the output - https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#keep-secret-data (note they use `True` but you should use `true`)

Also use `no_log: true` for the task below

tasks/presharedkey.yml
```diff
@@ -23,22 +38,22 @@
...
-      register: preshared_key_stat
+      register: __ha_cluster_preshared_key_stat
 
     - name: "Slurp existing {{ preshared_key_label }} from cluster nodes"
       slurp:
         src: "{{ preshared_key_dest }}"
-      register: preshared_key_slurp
-      when: preshared_key_stat.stat.exists
+      register: __ha_cluster_preshared_key_slurp
+      when: __ha_cluster_preshared_key_stat.stat.exists
```
richm, line 50: Same here - `no_log: true`

tasks/presharedkey.yml
```diff
@@ -23,22 +38,22 @@
...
 
-    - name: "Use the slurped {{ preshared_key_label }}"
+    - name: "Use the slurped {{ preshared_key_label }} from cluster nodes"
       set_fact:
         __ha_cluster_some_preshared_key: >
-          "{{ preshared_key_slurp.content }}"
-      when: preshared_key_stat.stat.exists
+          "{{ __ha_cluster_preshared_key_slurp.content }}"
+      when: __ha_cluster_preshared_key_stat.stat.exists
```
richm, line 56: Same here - `no_log: true`

richm:
@tomjelinek any ideas why we're getting the rhel-8-y failures?
```
TASK [linux-system-roles.ha_cluster : Install cluster packages] ****************
task path: /tmp/tmp98asuavs/tasks/install-and-configure-packages.yml:30
fatal: [/cache/rhel-8-y.qcow2]: FAILED! => {"changed": false, "failures": [], "msg": "
Depsolve Error occured:
Problem 1: conflicting requests
  - nothing provides corosynclib(x86-64) = 3.1.0-4.el8 needed by corosync-3.1.0-4.el8.x86_64
Problem 2: package pacemaker-2.0.5-9.el8.x86_64 requires corosync >= 2.0.0, but none of the providers can be installed
  - conflicting requests
  - nothing provides corosynclib(x86-64) = 3.1.0-4.el8 needed by corosync-3.1.0-4.el8.x86_64
", "rc": 1, "results": []}
```
We're using the `nightly/latest-RHEL-8/compose/HighAvailability` repo.

tomjelinek:
Several cluster components (including corosync) are split into multiple rmps. Some of those rpms go to rhel-appstream repo, while the others go to rhel-HighAvailability. Therefore, these two repos must point to the same release. If rhel-appstream is 'latest-RHEL-8.4', then rhel-HighAvailability must be 'latest-RHEL-8.4' as well.

jharuda:
[citest bad]

richm:
[citest pending]

jharuda:
[citest pending]

jharuda:
[citest pending]

tomjelinek:
[citest bad]

richm:
[citest pending]

richm:
[citest bad]

richm:
[citest pending]

PR #16, opened 2021-05-20
Code cleanup

tomjelinek:
* remove unnecessary quotes and backslashes
* readme readability improvements
* add missing spdx headers

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,12 +148,12 @@ See also:
 structure and default value:
 ```yaml
 ha_cluster_pcs_permission_list:
-  - type: "group"
-    name: "hacluster"
+  - type: group
+    name: hacluster
     allow_list:
-      - "grant"
-      - "read"
-      - "write"
+      - grant
+      - read
+      - write
 ```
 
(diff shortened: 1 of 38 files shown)


README.md
```diff
@@ -348,78 +348,78 @@ all:
...
+                value: 'true'
         operations:
-          - action: "start"
+          - action: start
             attrs:
-              - name: "timeout"
-                value: "30"
-          - action: "monitor"
+              - name: timeout
+                value: 30
```
richm, line 380: This changes the "type" from string to integer - where are these values used?  If the only place they are used is with `{{ value }}` then that should be ok - that will convert to string

tomjelinek: Yes, they are only used in `{{ value }}`. In general, they are all strings, though. So I fixed that and added an example and a test for setting timeout and intervals with time units.

tomjelinek:
[citest bad]

tomjelinek:
[citest bad]

richm (approved):
lgtm

richm:
Please use the Squash and merge button to merge this

PR #17, opened 2021-05-31
add pacemaker cluster properties configuration

tomjelinek:


README.md
```diff
@@ -174,6 +174,24 @@ string, default: `my-cluster`
 
 Name of the cluster.
 
+#### `ha_cluster_cluster_properties`
+
+structure, default: no properties
+
+```yaml
+ha_cluster_cluster_properties:
+  - attrs:
```
richm, line 183: So these are the attributes that are global, or cluster wide, or that apply to every host in the cluster?

What other elements might be in this array?
```yaml

      
    ha_cluster_cluster_properties:
      - attrs:
          - name: property1_name
            value: property1_value
          ...
      - somethingelse?
          - name: property1_name
            value: property1_value
```
?



    
    
      
          - name: property2_name




    
    
      
    <span class="blob-code-inner blob-code-marker" data-code-marker="+"><span class="pl-s1">        <span class="pl-ent">value</span>: <span class="pl-s">property2_value</span></span></span>

richm:
[citest bad]

README.md
tomjelinek: These attributes are cluster wide, they apply to the cluster itself. They are documented at https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/2.0/html/Pacemaker_Explained/s-cluster-options.html

There may be list of sets of properties. This is not supported yet, but it is a requested feature. Each set can have a rule which defines when the set applies and an id (name). I have not designed that in the details yet, the rough idea is:
```yaml
    ha_cluster_cluster_properties:
      - attrs:
          - name: property1_name
            value: property1_value
          - name: property2_name
            value: property2_value
      - attrs:
          - name: property1_name
            value: property1_value_for_weekend
        rule: date-spec weekdays=6-7
        id: weekend-properties
```
This sets value of `property1_name` to `property1_value_for_weekend` on weekends and to `property1_value` on Monday through Friday. Value of `property2_name` is constant.

richm: ok - so if a set of properties has no `rule` then the properties apply at all times to all nodes - and if there is no `id` then the implicit id is `default` or `global`

richm:
Any idea about the rhel-x (rhel-9) failure?
```
TASK [linux-system-roles.ha_cluster : Configure resource primitive 'xvm-no-attrs'] ***
task path: /tmp/tmp8vd56eue/tasks/pcs-cib-resource-create-pcs-0.10.yml:3
fatal: [/cache/rhel-x.qcow2]: FAILED! => {"changed": true, "cmd": ["pcs", "-f", "/tmp/ansible.u4p0cbh2_ha_cluster_cib_xml", "stonith", "create", "xvm-no-attrs", "fence_xvm"], "delta": "0:00:00.390779", "end": "2021-06-03 04:45:19.282228", "msg": "non-zero return code", "rc": 1, "start": "2021-06-03 04:45:18.891449", "stderr": "Error: Agent 'fence_xvm' is not installed or does not provide valid metadata: crm_resource: Metadata query for stonith:fence_xvm failed: No such device, Error performing operation: No such device, use --force to override", "stderr_lines": ["Error: Agent 'fence_xvm' is not installed or does not provide valid metadata: crm_resource: Metadata query for stonith:fence_xvm failed: No such device, Error performing operation: No such device, use --force to override"], "stdout": "", "stdout_lines": []}
```
Did something change for rhel-9?  Is there some other package that needs to be installed for rhel-9?

README.md
tomjelinek: > if a set of properties has no `rule` then the properties apply at all times to all nodes

yes

> if there is no `id` then the implicit id is `default` or `global`

implicit id is `cib-bootstrap-options` - it is set by pacemaker

tomjelinek:
I'm not sure what's happenning with rhel-9. The test fails due to `fence_xvm` fence agent missing. The test requests it to be installed:
```yaml
ha_cluster_fence_agent_packages:
      - fence-virt
```
It seems `fence-agents-all` is installed instead and `fence_xvm` is missing. I'll contact fence-agents maintainer.

tomjelinek:
There is a confirmed change in rhel-9. Fence-virt, which previously was a separate package, is being merged into fence-agents and will be provided as a subpackage. The tests were running in a middle of the transition, when fence-virt was not available. New version of the packages, which fix the issue, have been built today. The test should be passing again in a few days.

richm:
> There is a confirmed change in rhel-9. Fence-virt, which previously was a separate package, is being merged into fence-agents and will be provided as a subpackage. The tests were running in a middle of the transition, when fence-virt was not available. New version of the packages, which fix the issue, have been built today. The test should be passing again in a few days.

Is there a bz for that?

tomjelinek:
> Is there a bz for that?

Yes, [bz1965988](https://bugzilla.redhat.com/show_bug.cgi?id=1965988). And here is a PR ClusterLabs/fence-agents#390.

richm:
> > Is there a bz for that?
> 
> Yes, [bz1965988](https://bugzilla.redhat.com/show_bug.cgi?id=1965988). And here is a PR [ClusterLabs/fence-agents#390](https://github.com/ClusterLabs/fence-agents/pull/390).

Does that mean the `fence-virt` package will be dropped?

tomjelinek:
> > > Is there a bz for that?
> > 
> > 
> > Yes, [bz1965988](https://bugzilla.redhat.com/show_bug.cgi?id=1965988). And here is a PR [ClusterLabs/fence-agents#390](https://github.com/ClusterLabs/fence-agents/pull/390).
> 
> Does that mean the `fence-virt` package will be dropped?

No, `fence-virt` packages will be still available and `dnf install fence-virt` will work with no changes. It's merely going to be built from `fence-agents` rhpkg/fedpkg repos instead of `fence-virt` from now on.

tomjelinek:
[citest bad]

tomjelinek:
[citest bad]

richm:
[citest pending]

richm:
[citest bad]

richm:
@tomjelinek ready to merge?

PR #18, opened 2021-06-15
do not fail if openssl is not installed

tomjelinek:
OpenSSL is used in the role and automated tests to generate random
preshared keys. Previously, OpenSSL was a dependency of pcs. That is no
longer the case since pcs-0.10.8-2.

In the role, keys are no longer generated on the controller, so the role
does not install OpenSSL there. Instead, preshared keys are now
generated on target nodes and the role newly ensures OpenSSL is installed on
target nodes.

In tests, OpenSSL is needed on the controller to test cases when
preshared keys are distributed from the controller. Tests ensure OpenSSL
is installed on the controller.

### DIFF
--- a/tasks/presharedkey.yml
+++ b/tasks/presharedkey.yml
@@ -35,9 +35,17 @@
   delegate_to: localhost
 
 - name: Generate random {{ preshared_key_label }}
-  set_fact:
-    __ha_cluster_some_preshared_key: >
-      "{{ lookup('pipe', 'openssl rand -base64 {{ preshared_key_length }}') }}"
+  block:
+    - name: Generate {{ preshared_key_label }} using OpenSSL
+      command:
+        cmd: openssl rand -base64 {{ preshared_key_length | quote }}
+      register: __ha_cluster_openssl_rand
+      changed_when: no
+
+    - name: Fetch generated {{ preshared_key_label }}
+      set_fact:
+        __ha_cluster_some_preshared_key: >
+          "{{ __ha_cluster_openssl_rand.stdout }}"
   when:
     - not (preshared_key_src is string and preshared_key_src | length > 1)
   run_once: yes
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -17,6 +17,7 @@ __ha_cluster_fullstack_node_packages:
   - libknet1-plugins-all
   - pacemaker
   - pcs
+  - openssl  # used in the role for generating preshared keys
 
 __ha_cluster_services:
   - corosync
--- a/tests/tasks/fixture_psks.yml
+++ b/tests/tasks/fixture_psks.yml
@@ -1,7 +1,12 @@
 # SPDX-License-Identifier: MIT
 ---
-- name: Generate pre-shared keys and certificates on controller
+- name: Generate pre-shared keys and certificates on the controller
   block:
+    - name: Ensure OpenSSL is installed on the controller
+      package:
+        name: openssl
+        state: present
+
     - name: Ensure a directory for temporary files exists
       file:
         path: "{{ playbook_dir }}/tmp"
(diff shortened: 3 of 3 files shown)


vars/main.yml
```diff
@@ -17,6 +17,7 @@ __ha_cluster_fullstack_node_packages:
   - libknet1-plugins-all
   - pacemaker
   - pcs
+  - openssl  # used in the role for generating preshared keys
```
richm, line 20: You shouldn't pub `openssl` here unless `openssl` is also required on the *managed* nodes.  The usage in this PR is on the *control* node.

tomjelinek: When I drop `openssl` here, tests fail on my machines like this (I removed `no_log: true` so that the error message is visible):
```
TASK [linux-system-roles.ha_cluster : Generate fence_xvm.key using OpenSSL]
fatal: [rh84-node2]: FAILED! => {"changed": false, "cmd": "openssl rand -base64 512", "msg": "[Errno 2] No such file or directory: b'openssl': b'openssl'", "rc": 2, "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```
`rh84-node2` is a managed node, controller node is `rh84-node1`. My intention was to use `openssl` on managed nodes in task `Generate {{ preshared_key_label }} using OpenSSL` in `tasks/presharedkey.yml`, this very pull request adds that. Maybe I am missing something here?

richm: > Maybe I am missing something here?

No, my bad. I thought you were only using openssl on the control node.

richm:
[citest bad]

spetrosi:
Seems like the checks fail because of the move to BaseOS CI and can be ignored for now, @richm please  verify.
Otherwise, lgtm.

richm:
[citest bad]

richm (approved):
lgtm - fedora 34 failures should be fixed now - retesting

spetrosi:
I have noticed that the openssl package returned to the test images. I've checked couple of tests and the `Ensure OpenSSL is installed on the controller` task reports `Nothing to do`. @tomjelinek could you please verify this?

tasks/presharedkey.yml
```diff
@@ -35,9 +35,17 @@
...
 - name: Generate random {{ preshared_key_label }}
-  set_fact:
-    __ha_cluster_some_preshared_key: >
-      "{{ lookup('pipe', 'openssl rand -base64 {{ preshared_key_length }}') }}"
+  block:
+    - name: Generate {{ preshared_key_label }} using OpenSSL
+      command:
+        cmd: openssl rand -base64 {{ preshared_key_length | quote }}
+      register: __ha_cluster_openssl_rand
+      changed_when: no
```
richm, line 43: add `no_log: true` here

tasks/presharedkey.yml
```diff
@@ -35,9 +35,17 @@
...
+
+    - name: Fetch generated {{ preshared_key_label }}
+      set_fact:
+        __ha_cluster_some_preshared_key: >
+          "{{ __ha_cluster_openssl_rand.stdout }}"
   when:
     - not (preshared_key_src is string and preshared_key_src | length > 1)
   run_once: yes
```
richm, line 51: add `no_log: true` here - probably should have had it in the previous version

tomjelinek:
> I have noticed that the openssl package returned to the test images. I've checked couple of tests and the `Ensure OpenSSL is installed on the controller` task reports `Nothing to do`. @tomjelinek could you please verify this?

I read all the logs from CI and found four which installed the openssl package:
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-18-0021093-centos-8-20210615-091100/artifacts/test.log.html
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-18-0021093-fedora-33-20210615-091408/artifacts/test.log.html
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-18-0021093-rhel-8-y-20210615-091724/artifacts/test.log.html
https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-18-0021093-rhel-8-20210615-105214/artifacts/test.log.html

Anyway, I'd rather be on the safe side and make the role install all packages it depends on.

tasks/presharedkey.yml
tomjelinek: `no_log: true` is set for the whole block containing this task. I went through all logs from CI and verified that output of this task is indeed hidden.

tasks/presharedkey.yml
tomjelinek: `no_log: true` is set for the whole block containing this task. I went through all logs from CI and verified that output of this task is indeed hidden.

spetrosi (approved):
lgtm

richm (approved):
lgtm - ready to merge

PR #19, opened 2021-08-04
use correct python-cryptography package

richm:
Use `python-cryptography` on EL7 - use `python-cryptography3` everywhere
else
https://bugzilla.redhat.com/show_bug.cgi?id=1989962


nhosoi (approved):
lgtm

richm:
well, since my fixes just break it instead, I'm withdrawing this PR until we actually do have a problem

PR #21, opened 2021-08-31
Use the openssl command-line interface instead of the openssl module

nhosoi:
in the test helper task tests/tasks/fixture_psks.yml.

This is to avoid using the non ansible-core module.


richm:
[citest bad]

richm (approved):
lgtm

PR #23, opened 2021-09-08
fix password_hash salt length

richm:
Ansible 2.11 has strict checking for password_hash salt length.
If the given salt length is not equal to, or greater than, the
salt length for the hash algorithm (depending on the algorithm, the
salt length must be equal, or less than), Ansible will emit an
error like "invalid salt size".  To fix this, use the `truncate`
filter to ensure the salt length is not too long.


nhosoi (approved):
lgtm

Issue #29, opened 2021-11-04
Add support for SBD devices


stdevel:
In `tasks/cluster-enable-disable.yml` there is a comment that SBD is currently disabled because it is not supported, yet:

```
# The role does not support configuring SBD yet, therefore we always disable it
```

Are there plans to support it in the future? I'm using this role to setup RHEL HA and the official documentation mentions that SBD is supported by Red Hat.

richm:
@tomjelinek what are the plans for SBD support?

tomjelinek:
Adding support for SBD to the system role is a high priority task. It has not made it to the top of the TODO list yet, however.

stdevel:
How do you plan to implement it? I implemented some `post_tasks` as workaround, maybe I can contribute?

stdevel:
I felt free to show you what I already got in merge request #31. It is not completed, yet (see details in the PR) - but it already works for my environment.

richm:
fixed by https://github.com/linux-system-roles/ha_cluster/pull/31

PR #31, opened 2021-11-10
add SBD support

stdevel:
As discussed in #29 I started implementing my workarounds in order to support SBD.

This is **not** considered as *ready to merge*, yet. I was able to use this code in order to configure SBD on a 2-node cluster using the following code:

```yaml
- name: linux-system-roles.ha_cluster
  ha_cluster_cluster_name: "{{ cluster_name }}"
  ha_cluster_hacluster_password: "{{ cluster_password }}"
  ha_cluster_resource_primitives:
    - id: fence-sbd
      agent: 'stonith:fence_sbd'
      instance_attrs:
        - attrs:
            - name: devices
              value: /dev/sdb
  ha_cluster_resource_groups: "{{ cluster_groups }}"
  ha_cluster_sbd_enable: true
  ha_cluster_sbd_devices:
    - /dev/sdb
  ha_cluster_sbd_startmode: "clean"
```

Currently, there are still two TODOs:

- [x] make creating SBD headers on disk idempotent
- [ ] automatically configure SBD disks as PCS fence resources - currently this needs to be done manually, e.g. using the `ha_cluster_resource_primitives` variable (see above)

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,4 +1,3 @@
 # SPDX-License-Identifier: MIT
 ---
-- name: Load softdog module
-  command: modprobe softdog
+# Put handlers for tasks here.
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,3 +1,4 @@
 # SPDX-License-Identifier: MIT
 ---
-# Put handlers for tasks here.
+- name: Load softdog module
+  shell: modprobe softdog
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -1,4 +1,4 @@
 # SPDX-License-Identifier: MIT
 ---
 - name: Load softdog module
-  shell: modprobe softdog
+  command: modprobe softdog
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -25,7 +25,7 @@
 
 - name: Enable softdog module
   copy:
-    content: "softdog"
+    content: softdog
     dest: /etc/modules-load.d/softdog.conf
     owner: root
     group: root
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -10,6 +10,7 @@
 #   when:
 #     - ha_cluster_sbd_devices
 
+# TODO: deadlock if SBD was working before and cluster was stopped
 - name: Create SBD devices
   command: sbd -d {{ item }} {{ ha_cluster_sbd_create_params }} create
   run_once: true
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -1,22 +1,22 @@
 # SPDX-License-Identifier: MIT
 ---
-# TODO: make creating SBD headers idempotent
-# - name: Check SBD device headers
-#   command: "sbd -d {{ item }}  dump"
-#   run_once: true
-#   ignore_errors: true
-#   register: "sbd_dump_{{ item }}"
-#   loop: "{{ ha_cluster_sbd_devices }}"
-#   when:
-#     - ha_cluster_sbd_devices
-
-# TODO: deadlock if SBD was working before and cluster was stopped
-- name: Create SBD devices
-  command: sbd -d {{ item }} {{ ha_cluster_sbd_create_params }} create
+- name: Check SBD device headers
+  command: "sbd -d {{ item }} dump"
   run_once: true
+  ignore_errors: true
+  changed_when: false
+  register: "sbd_dump"
   loop: "{{ ha_cluster_sbd_devices }}"
   when:
     - ha_cluster_sbd_devices
+
+- name: Create SBD devices
+  command: sbd -d {{ item.item }} {{ ha_cluster_sbd_create_params }} create
+  loop: "{{ sbd_dump['results'] }}"
+  when:
+    - item.rc == 1
+    - ha_cluster_sbd_devices
+  run_once: true
 
 - name: Enable softdog module
   copy:
@@ -36,11 +36,10 @@
     mode: '0644'
   when: ha_cluster_sbd_devices
 
-- name: Enable and start SBD
+- name: Enable SBD
   service:
     name: sbd
     enabled: true
-    state: started
 
 # TODO: automatically create stonith fence resources
 #       e.g. via pcs-cib-resource-create-pcs-0.10.yml?
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -26,18 +26,13 @@
     mode: '0644'
   notify: Load softdog module
 
-- name: Set SBD parameters
-  replace:
-    path: /etc/sysconfig/sbd
-    regexp: "{{ item.regexp }}"
-    replace: "{{ item.replace }}"
-  loop:
-    - regexp: '^(.*)SBD_DEVICE=(.*)$'
-      replace: 'SBD_DEVICE="{{ ha_cluster_sbd_devices | join(";") }}"'
-    - regexp: '^(.*)SBD_OPTS=(.*)$'
-      replace: SBD_OPTS="{{ ha_cluster_sbd_daemon_params }}"
-    - regexp: '^(.*)SBD_STARTMODE=(.*)$' 
-      replace: "SBD_STARTMODE={{ ha_cluster_sbd_startmode }}"
+- name: Create SBD configuration file
+  template:
+    src: templates/sbd.j2
+    dest: /etc/sysconfig/sbd
+    owner: root
+    group: root
+    mode: '0644'
   when: ha_cluster_sbd_devices
 
 - name: Enable and start SBD
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -11,7 +11,7 @@
 #     - ha_cluster_sbd_devices
 
 - name: Create SBD devices
-  command: "sbd -d {{ item }} {{ ha_cluster_sbd_create_params }} create"
+  command: sbd -d {{ item }} {{ ha_cluster_sbd_create_params }} create
   run_once: true
   loop: "{{ ha_cluster_sbd_devices }}"
   when:
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -1,45 +1,88 @@
 # SPDX-License-Identifier: MIT
 ---
-- name: Check SBD device headers
-  command: "sbd -d {{ item }} dump"
-  run_once: true
-  ignore_errors: true
-  changed_when: false
-  register: "sbd_dump"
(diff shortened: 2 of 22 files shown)


handlers/main.yml
```diff
@@ -1,3 +1,4 @@
 # SPDX-License-Identifier: MIT
 ---
-# Put handlers for tasks here.
+- name: Load softdog module
+  shell: modprobe softdog
```
richm, line 4: There is a `modprobe` module - https://docs.ansible.com/ansible/latest/collections/community/general/modprobe_module.html#ansible-collections-community-general-modprobe-module
This will unfortunately introduce a dependency on community.general in meta/requirements.yml

or - I suggest using `command` instead of `shell`

Is `softdog` present on all EL7+ and Fedora platforms?

tasks/sbd.yml
```diff
@@ -0,0 +1,61 @@
...
+#   register: "sbd_dump_{{ item }}"
+#   loop: "{{ ha_cluster_sbd_devices }}"
+#   when:
+#     - ha_cluster_use_sbd
+#     - ha_cluster_sbd_devices
+#   tags:
+#     - sbd
+
+- name: Create SBD devices
+  command: "sbd -d {{ item }} {{ ha_cluster_sbd_create_params }} create"
```
richm, line 17: you don't need double quotes here

tasks/sbd.yml
```diff
@@ -0,0 +1,61 @@
...
+  loop: "{{ ha_cluster_sbd_devices }}"
+  when:
+    - ha_cluster_use_sbd
+    - ha_cluster_sbd_devices
+  tags:
+    - sbd
+
+- name: Enable softdog module
+  copy:
+    content: "softdog"
```
richm, line 28: you don't need double quotes here

tasks/sbd.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Create SBD devices
+  command: "sbd -d {{ item }} {{ ha_cluster_sbd_create_params }} create"
+  run_once: true
+  loop: "{{ ha_cluster_sbd_devices }}"
+  when:
+    - ha_cluster_use_sbd
+    - ha_cluster_sbd_devices
+  tags:
+    - sbd
```
richm, line 24: What will this tag be used for?

tasks/sbd.yml
```diff
@@ -0,0 +1,61 @@
...
+    path: /etc/sysconfig/sbd
+    regexp: "{{ item.regexp }}"
+    replace: "{{ item.replace }}"
+  loop:
+    - regexp: '^(.*)SBD_DEVICE=(.*)$'
+      replace: 'SBD_DEVICE="{{ ha_cluster_sbd_devices | join(";") }}"'
+    - regexp: '^(.*)SBD_OPTS=(.*)$'
+      replace: SBD_OPTS="{{ ha_cluster_sbd_daemon_params }}"
+    - regexp: '^(.*)SBD_STARTMODE=(.*)$' 
+      replace: "SBD_STARTMODE={{ ha_cluster_sbd_startmode }}"
```
richm, line 48: Is there a chance that some other software will be managing the contents of `/etc/sysconfig/sbd` e.g. do you want to handle the case where the user may have added something to `SBD_DEVICE`, `SBD_OPTS`, or `SBD_STARTMODE`, and you want to preserve it, in addition to the changes made by the ha_cluster role? If so, then I would suggest writing a module in order to handle preserving existing values.

If not, then I would suggest creating a jinja2 `template` for this.

handlers/main.yml
stdevel: Thanks for pointing out - I would also suggest not using the module as it's just a simple task and adding another dependency for this doesn't look senseful to me. Will use `command` instead.
`softdog` is available on Fedora and EL7 and newer - it's part of the kernel.

tasks/sbd.yml
stdevel: Are you sure? Seems not to work for me:

```
FAILED! => {"msg": "The conditional check 'ha_cluster_use_sbd' failed. The error was: error while evaluating conditional (ha_cluster_use_sbd): 'ha_cluster_use_sbd' is undefined
```

tasks/sbd.yml
stdevel: Good point!

tasks/sbd.yml
stdevel: They were used for testing purposes - I will remove them.

tasks/sbd.yml
stdevel: Good point!

tasks/sbd.yml
stdevel: I wouldn't expect the user to alter the SBD configuration outside this role. It is mainly configured for cluster purposes. Also, the user has the option to set the most configured values using the role variables.
I will convert this into a jinja2 `template`.

tomjelinek:
Thank you for the pull request. I should be able to go through it in a couple of weeks, once I finish tasks with higher priority.

stdevel:
Thanks for the feedback @tomjelinek. No worries, I will also require some time to implement the last two features. :-)

richm:
[citest commit:0b4bda3955f78264d4cbcb9c8bfd7d51db8350ae]

tomjelinek:
@stdevel Thank you for your pull request. I went through it, added more options, fixed a few issues and added automated tests.

stdevel:
Thanks for the review @tomjelinek - happy to help!
Do you have any ideas on how to implement the last remaining step (see first comment)?

Best wishes,
Christian.

tomjelinek:
[citest]

tomjelinek:
[citest]

tomjelinek:
@stdevel For now, I would settle with manually configuring the fence resources. We haven't even implemented corresponding functionality in pcs yet, which is pretty much a prerequisite. Once we analyze how it works and design and implement it in pcs, I'd move to add it to the role. Hopefully, going from pcs to the role would be a straightforward process.

tomjelinek:
- Fedora
  - `tests_cib_constraints_create` fails due to a bug in pcs. New pcs packages will be build to fix this.
  -  Rest of the failures is due to `Failed to download packages: Cannot download {package}: All mirrors were tried`
- RHEL and CentOS
  -  All tests fail on the first task with error `/bin/sh: /usr/bin/python: No such file or directory` and `The module failed to execute correctly, you probably need to set the interpreter.`

Is the CI broken or do I need to fix something in the role?

richm:
> * Fedora
>       
>       * `tests_cib_constraints_create` fails due to a bug in pcs. New pcs packages will be build to fix this.
>       * Rest of the failures is due to `Failed to download packages: Cannot download {package}: All mirrors were tried`

Probably some sort of transient network issue.  Try these again.

> 
>     * RHEL and CentOS
>       
>       * All tests fail on the first task with error `/bin/sh: /usr/bin/python: No such file or directory` and `The module failed to execute correctly, you probably need to set the interpreter.`
> 
> 
> Is the CI broken or do I need to fix something in the role?

CI is broken and being worked on.

richm:
[citest commit:60ac893ea02ae4d3d5cd796d88f21bf98aafa58b]

tomjelinek:
[citest bad]

richm:
[citest bad]

richm:
[citest pending]

richm:
@tomjelinek 1) ignore the CI failures of the tft bot 2) I "kicked" some worker pods and now it seems faster 3) the rhel 8, rhel 9, and fedora failures are due to a pcs bug?

tomjelinek:
> 3) the rhel 8, rhel 9, and fedora failures are due to a pcs bug?

Yes

richm:
[citest]

richm:
[citest commit:2b6e2f36508726261649bb04b7dc82ddeaf93d0b]

tomjelinek:
[citest]

tomjelinek:
[citest]

tomjelinek:
[citest]

richm:
@tomjelinek still working on the tft-bot CI - so if you are seeing failures there, you can ignore them.  I'm more interested in the systemroller CI jobs - if those pass, then we're good to go.

tomjelinek:
@richm Ah, ok. Thanks for the heads-up. How do I get to run those systemroller CI jobs, though?

richm:
[citest commit:74fbd6b43c33d2673cc02d967d2147add80aa01b]

tomjelinek:
[citest commit:42d4a6cdfceb6685a015a35de21fa0406d52d93e]

richm:
[citest commit:42d4a6cdfceb6685a015a35de21fa0406d52d93e]

tomjelinek:
[citest commit:14f1f504b74b8a616aa331158ae90b19fa6cd5cc]

tomjelinek:
[citest commit:14f1f504b74b8a616aa331158ae90b19fa6cd5cc]

richm:
[citest commit:14f1f504b74b8a616aa331158ae90b19fa6cd5cc]

richm:
@tomjelinek this looks good to me - the CI failures are known e.g. https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/linux-system-roles-ha_cluster-pull-linux-system-roles_ha_cluster-31-14f1f50-rhel-8-20220216-162621/artifacts/ansible.log
```
TASK [linux-system-roles.ha_cluster : Configure location constraint '1'] *******
task path: /tmp/tmprryznki0/tests/roles/linux-system-roles.ha_cluster/tasks/pcs-cib-constraint-location.yml:3
Wednesday 16 February 2022  16:28:19 +0000 (0:00:00.843)       0:01:29.261 **** 
fatal: [/cache/rhel-8.qcow2]: FAILED! => {
    "changed": true,
    "cmd": [
        "pcs",
        "-f",
        "/tmp/ansible.2ocxjofa_ha_cluster_cib_xml",
        "--",
        "constraint",
        "location",
        "regexp%d\\d+",
        "prefers",
        "node1=INFINITY"
    ],
    "delta": "0:00:00.464496",
    "end": "2022-02-16 11:28:18.529249",
    "rc": 1,
    "start": "2022-02-16 11:28:18.064753"
}

STDOUT:

Warning: Validation for node existence in the cluster will be skipped


STDERR:

Error: Unable to update cib
...
Call failed: Update does not conform to the configured schema
```
I think you said this was a known failure in pcs in rhel8?  So I think this PR is ready to go.

tomjelinek:
> I think you said this was a known failure in pcs in rhel8?

Yes. New pcs packages fixing this issue have already been built and are queued to be released.

PR #36, opened 2022-02-01
fix default pcsd permissions

tomjelinek:
Previously, permissions were set for 'hacluster' group. Correct name of the group is 'haclient'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ structure and default value:
 ```yaml
 ha_cluster_pcs_permission_list:
   - type: group
-    name: hacluster
+    name: haclient
     allow_list:
       - grant
       - read
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -24,7 +24,7 @@ ha_cluster_cluster_name: my-cluster
 
 ha_cluster_pcs_permission_list:
   - type: group
-    name: hacluster
+    name: haclient
     allow_list:
       - grant
       - read


tomjelinek:
> If a customer already has ha_cluster deployed, will this change the existing deployment to use haclient?

If the customer has a playbook calling `ha_cluster` role and `ha_cluster_pcs_permission_list` variable is not specified, then running the playbook will change the existing deployment to use `haclient`.
If they specify `ha_cluster_pcs_permission_list`, then its value will be used and the change done by this commit doesn't matter.

richm (approved):
lgtm

tomjelinek:
It seems there are still issues with CI, tests seems to be failing outside of the role control:
* ERROR: Inventory is empty, tests did not run
* Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
* Failed to connect to the host via ssh: ssh: connect to host 127.0.0.3 port 3426: Connection refused

richm:
> It seems there are still issues with CI, tests seems to be failing outside of the role control:
> 
>     * ERROR: Inventory is empty, tests did not run
> 
>     * Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
> 
>     * Failed to connect to the host via ssh: ssh: connect to host 127.0.0.3 port 3426: Connection refused

yes - I'm investigating all of these - just ignore them and merge your fix

richm:
@tomjelinek how urgent is this?  Do we need a BZ for this to get into RHEL asap?

tomjelinek:
@richm I don't think this is urgent priority. There is a workaround. However, it would be nice if this was included in the upcoming release.

PR #42, opened 2022-03-11
add support for advanced corosync configuration

tomjelinek:


tasks/check-and-prepare-role-variables.yml
```diff
@@ -36,3 +36,28 @@
...
+        and not ha_cluster.sbd_devices | default([])
+        and __ha_cluster_all_node_names | length is even
+        and not __ha_cluster_qdevice_in_use
+      }}"
+
+- name: Fail if SBD needs ATB enabled and the user configured ATB to be disabled
+  fail:
+    msg: Cannot set auto_tie_breaker to disabled when SBD needs it to be enabled
+  when:
+    - __ha_cluster_sbd_needs_atb
```
richm, line 59: you can't have a "bare" variable here - so use `__ha_cluster_sbd_needs_atb | bool`

tasks/check-and-prepare-role-variables.yml
```diff
@@ -36,3 +36,28 @@
...
+    - >
```
richm, line 60: `when` conditions don't need to be strings - so you can do
```
  - ha_cluster_quorum.options | default([])
    | selectattr('name', 'match', '^auto_tie_breaker$')
    | map(attribute='value') | select('in', ['0', 0]) | list | length > 0

File: tests/tasks/assert_corosync_config.yml [hunk 3]
```diff
@@ -0,0 +1,24 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Fetch corosync configuration from the cluster
+  # Remove the contents of 'Nodes:' section as it depends on
+  # testing environment
+  shell:
+    cmd: pcs cluster config | sed -e '/^Nodes:$/,/^[^ ]/ { /^ /d }'
```
richm, line 7: `shell` should be written like this:
```yaml
shell: |
  set -euo pipefail
  pcs cluster config | sed -e '/^Nodes:$/,/^[^ ]/ { /^ /d }'
```

tests/tasks/assert_corosync_config.yml
```diff
@@ -0,0 +1,24 @@
...
+  register: __test_pcs_corosync_config
+  changed_when: no
+
+- name: Print real corosync configuration
+  debug:
+    var: __test_pcs_corosync_config.stdout_lines
```
richm, line 13: if you want users to be able to hide this, I suggest using `verbosity`: https://docs.ansible.com/ansible/2.9/modules/debug_module.html#debug-module
e.g. using `verbosity: 2` means that the user would need to run ansible with `-vv` in order to see this output

however, if you think that this should always be in the output, that's fine too - we have had users complain about the amount of output in some of our roles - the storage role team recently merged a change to reduce the amount of output by default - https://github.com/linux-system-roles/storage/pull/256

tests/tasks/assert_corosync_config.yml
```diff
@@ -0,0 +1,24 @@
...
+
+- name: Print expected corosync configuration
+  debug:
+    var: __test_expected_lines | select | list
+
+- name: Check corosync configuration
+  assert:
+    that:
+      - >
```
richm, line 22: `when` and `that` conditions do not have to be strings - they can be "raw" jinja code - so
```
that:
  - __test_pcs_corosync_config.stdout_lines
    == __test_expected_lines | select | list
```

tests/tasks/assert_corosync_config.yml
richm: nevermind - this is in test code, so yes, you always want to see this

tests/tests_cluster_advanced_knet_full.yml
```diff
@@ -0,0 +1,110 @@
...
+              - '  auto_tie_breaker: 1'
+              - '  wait_for_all: 1'
+
+        - name: Check cluster status
+          include_tasks: tasks/assert_cluster_running.yml
+
+      always:
+        - name: Unset node addresses variable
+          set_fact:
+            ha_cluster:
```
richm, line 109: @nhosoi does this work to "unset" a global variable?  Or does it set the value to `None`?

tests/tests_sbd_needs_atb_while_atb_disabled.yml
```diff
@@ -0,0 +1,53 @@
...
+              sbd_devices: []
+
+        - block:
+            - name: Run the role
+              include_role:
+                name: linux-system-roles.ha_cluster
+          rescue:
+            - name: Check errors
+              assert:
+                that: "{{
```
richm, line 33: you can use
```
that: ansible_failed_result.msg ==
        'Cannot set auto_tie_breaker to disabled when SBD needs it ' ~
        'to be enabled'
```
`~` is the jinja string concat operator - https://jinja.palletsprojects.com/en/2.9.x/templates/#other-operators

richm:
The yamllint errors are in the converted collection format - looks like there is a bug with the list-of-list processing in our converter - this is what it is complaining about:
```yaml
      links:
        -   - name: transport
```
there are two spaces between the `-`

tasks/check-and-prepare-role-variables.yml
tomjelinek: Ack, fixed

tasks/check-and-prepare-role-variables.yml
tomjelinek: Right, fixed

tests/tasks/assert_corosync_config.yml
tomjelinek: Fixed, thanks!

README.md
```diff
@@ -194,13 +194,11 @@ ha_cluster_transport:
     - name: option2_name
       value: option2_value
   links:
-    -
-      - name: option1_name
+    - - name: option1_name
```
richm, line 197: ok - if this doesn't work, then just disable yamllint hyphen checking
```yaml
  -
    - name: option1_name  # yamllint disable-line rule:hyphens
```
that way when it is converted, it will come out like this:
```
  -   - name: option1_name  # yamllint disable-line rule:hyphens
```
and yamllint will not complain

tests/tasks/assert_corosync_config.yml
tomjelinek: Ack, fixed

README.md
richm: ok - looks like using `- - name: ...` didn't work - so use my suggestion

tests/tests_sbd_needs_atb_while_atb_disabled.yml
tomjelinek: Yes, that's definitely better, fixed. Thanks!

README.md
tomjelinek: done

richm:
[citest commit:36230ea967a7a160139e5ab3cb73a8a2558310c2]

richm:
lgtm - just waiting for the `ansible-2.12` tests

richm (approved):
lgtm

PR #43, opened 2022-03-31
add support for configuring bundle resources

tomjelinek:


tomjelinek:
[citest commit:2621afd58d126e269197648d0da13d05b0e2b30a]

richm:
> [citest commit:2621afd58d126e269197648d0da13d05b0e2b30a]

CI is down - big outage

richm:
> > CI is down - big outage
> 
> Ah, ok. Thanks for the heads up. Will it run tests automatically once it comes back online or do I need to start them manually?

I will start them

tests/tests_cib_resources_create.yml
```diff
@@ -188,6 +202,63 @@
                 value: a value
       - resource_id: group3
 
+    ha_cluster_resource_bundles:
+      - id: bundle-minimal
+        container:
+          type: docker
```
richm, line 208: can you use `podman` for this?
Do you need to install `podman` on the managed hosts?
Do you need to pull the image `my:image1` to the managed hosts?
Do you need to configure the firewall to allow the specified ports?

tomjelinek: In this test, the cluster is configured not to start any resources. Therefore, there's no need to install `podman` (or `docker`) on the hosts, providing images and allowing ports in firewall. This makes the test more independent on the environment it runs in, while having no effect on the test's ability to verify that the role works correctly.

Yes, `podman` can be used instead of `docker`. However, it doesn't actually change anything, as the container won't be started by the cluster anyway. And if the role is to be made capable of managing RHEL 7 hosts, we must stick to `docker`, as `podman` is not supported on RHEL 7.

tomjelinek:
rhel-8 failures are caused by not using the latest pcs version. Fixed pcs versions are pcs-0.10.10-4.el8_5.1 and pcs-0.10.12-3.el8

tests/tests_cib_resources_create.yml
richm: > In this test, the cluster is configured not to start any resources. Therefore, there's no need to install `podman` (or `docker`) on the hosts, providing images and allowing ports in firewall. This makes the test more independent on the environment it runs in, while having no effect on the test's ability to verify that the role works correctly.
> 
> Yes, `podman` can be used instead of `docker`. However, it doesn't actually change anything, as the container won't be started by the cluster anyway.

How is the container pulled/started/managed?  What are the requirements for using this feature?

> And if the role is to be made capable of managing RHEL 7 hosts, we must stick to `docker`, as `podman` is not supported on RHEL 7.

ack

tomjelinek: > How is the container pulled/started/managed? What are the requirements for using this feature?

With this patch, the system role will be able to configure pacemaker to run (start, stop, migrate) containerized resources (a.k.a. bundles). In order for it to actually work, the appropriate container launch technology (docker, podman or rkt) must be installed, and a fully configured container image must be supplied on every node allowed to run the bundle. Also, a resource agent for the container launcher must be installed on the nodes. Currently, it is installed automatically as a dependency of pacemaker packages.

Both installing a container launcher and building / distributing container images is out of scope of HA cluster management tools. It is possible to install a container launcher, or any other package, via the role by specifying it in `ha_cluster_extra_packages` variable. There is no support for managing container images at the moment, though.

I guess this should be documented in the readme file, right? That is, if the role itself is not required to be able to build container images.

richm: > > How is the container pulled/started/managed? What are the requirements for using this feature?
> 
> With this patch, the system role will be able to configure pacemaker to run (start, stop, migrate) containerized resources (a.k.a. bundles). In order for it to actually work, the appropriate container launch technology (docker, podman or rkt) must be installed, and a fully configured container image must be supplied on every node allowed to run the bundle. Also, a resource agent for the container launcher must be installed on the nodes. Currently, it is installed automatically as a dependency of pacemaker packages.
> 
> Both installing a container launcher and building / distributing container images is out of scope of HA cluster management tools. It is possible to install a container launcher, or any other package, via the role by specifying it in `ha_cluster_extra_packages` variable. There is no support for managing container images at the moment, though.

We are working on a `podman` role - https://github.com/linux-system-roles/podman/

> 
> I guess this should be documented in the readme file, right? That is, if the role itself is not required to be able to build container images.

Right.  It would be nice to know what requirements/steps are needed to manage the container images - perhaps this is something we could add to the `podman` role (but then users of `docker` and other container technologies would be on their own).  Then the `ha_cluster` role could use the `podman` role to do that management.

richm (approved):
lgtm

PR #49, opened 2022-05-13
fix decoding variables from an Ansible vault

tomjelinek:
Workaround Ansible issue https://github.com/ansible/ansible/issues/24425
Before fix, the role was failing with the following message:
object of type 'AnsibleVaultEncryptedUnicode' has no len()


tomjelinek:
I'd like to write a test for this issue. How do I instruct CI to decrypt a vault and how do I pass a vault password? Does the CI support it?

richm:
> I'd like to write a test for this issue. How do I instruct CI to decrypt a vault and how do I pass a vault password? Does the CI support it?

Not sure how we would do this in CI.  How are you testing manually?

tomjelinek:
When testing manually, I run a playbook with the `--ask-vault-pass` option and I enter a password once prompted.

richm:
> When testing manually, I run a playbook with the `--ask-vault-pass` option and I enter a password once prompted.
> 
> The playbook contains a variable encrypted with `ansible-vault encrypt_string 'your-hacluster-password' --name ha_cluster_hacluster_password`. Encrypting the variable and putting it into the playbook is a one time action.

Have you tried it with `--vault-password-file pwd_file`?

Perhaps CI could do something like this:
Each test that uses vault encrypted data has a file `tests/tests_name_of_test.inventory` (containing vars with vault encrypted data) and a file `tests/tests_name_of_test.vault_pwd` (containing the password used to encrypt/decrypt the vault encrypted data for the test) - if CI sees such files, then it changes the ansible-playbook command line to
```
ansible-playbook -i /path/to/tests/tests_name_of_test.inventory -i ... --vault-password-file /path/to/tests/tests_name_of_test.vault_pwd .... tests/tests_name_of_test.yml
```

tomjelinek:
That sounds promising. I'll try to write a test like that and run it on my machine.

richm:
> Based on my testing, it seems the issue is not triggered when variables with vault-encrypted data are present in an inventory. It is, however, triggered when such variables are present in a playbook or a vars file. Vars file can be passed to ansible-playbook like this: `ansible-playbook -i ... --vault-password-file /path/to/tests/tests_name_of_test.vault_pwd -e @/path/to/tests/tests_name_of_test.vars tests/tests_name_of_test.yml`.
> 
> Could the CI be set to detect just `tests/tests_name_of_test.vault_pwd` files and pass them to `ansible-playbook` command?

Yes.  Instead of passing in `-i /path/to/tests/tests_name_of_test.inventory`, CI would use `-e @/path/to/tests/tests_name_of_test.vars` - that should work

richm:
ok - lgtm - and this will require CI changes for automated testing.

tomjelinek:
Should I wait for the CI to get updated and the new test pass before merging?

richm:
> Should I wait for the CI to get updated and the new test pass before merging?

Yes, please - it will be a couple of days before CI is fixed.

tomjelinek:
> > Should I wait for the CI to get updated and the new test pass before merging?
> 
> Yes, please - it will be a couple of days before CI is fixed.

Any updates on adding support for ansible vault tests?

richm:
https://gitlab.com/testing-farm/gluetool-modules/-/merge_requests/205 - MR to implement https://github.com/linux-system-roles/ha_cluster/pull/49#issuecomment-1224497673 in CI

tomjelinek:
Sorry for the delayed response. The described solution looks perfect to me. I'll update the PR accordingly.

richm:
> We'll also add support for tox-lsr for this, so that you can simply run tox -e qemu-ansible-core-2.13 -- --image-name centos-9 tests/tests_sometest.yml and it will automatically use the vault password and the vault variables/no-vault-variables.txt

This is now done in the latest version of tox-lsr https://github.com/linux-system-roles/tox-lsr/commit/d7968926eac6a75b1f36c04ae7624f968256f071

richm (approved):
just waiting for CI to be updated with vault support

richm:
[citest]

spetrosi:
@tomjelinek are failing CentOS-Stream-8 tests expected?

richm:
> @tomjelinek are failing CentOS-Stream-8 tests expected?

Not sure, but it is really strange that only centos stream 8 has the problem with the expected key not matching the actual key.  will run again to see if this is a flake

[citest bad]

richm:
> > @tomjelinek are failing CentOS-Stream-8 tests expected?
> 
> Not sure, but it is really strange that only centos stream 8 has the problem with the expected key not matching the actual key. will run again to see if this is a flake
> 
> [citest bad]

The failures are a consequence of the move to baseos CI, which runs tests *in parallel* and *uses the same git clone for every test*.  There is a subtle bug in the ha_cluster tests - the `*_psks.yml` tests generate new secrets in the clone tests/tmp directory[1].  Which means one of the `*_psks.yml` tests can overwrite the values used by the other test, and get errors.  This also explains why sometimes you get failures in one test but not the other - it depends on the order they are run.  This worked in the legacy CI test environment because the tests were run sequentially.

[1] @tomjelinek the secrets are already cloned into the tests/tmp directory and checked into git, but the tests also recreate these values?

richm:
https://github.com/linux-system-roles/ha_cluster/pull/65 - PR to fix the temp files being overwritten issue

tomjelinek:
@richm Thanks for fixing this tmp directory issue.

I don't think the tests/tmp is checked into git. I don't see it in GitHub and it is ignored in https://github.com/linux-system-roles/ha_cluster/blob/master/tests/.gitignore#L3

PR #60, opened 2022-08-01
replace yes, no, default with true, false, d

tomjelinek:


richm:
[citest]

richm (approved):

tomjelinek:
All CI failures are due to tasks in `tests/tasks/fixture_psks.yml`.

```yaml
    - name: List packages on the controller to see if OpenSSL is installed
      package_facts:
      delegate_to: localhost
      run_once: true
````
Fails with: Could not detect a supported package manager from the following list: ['pkg', 'portage', 'apt', 'rpm'], or the required Python library is not installed. Check warnings for details.

```yaml
    - name: Ensure OpenSSL is installed on the controller
      package:
        name: openssl
        state: present
      when: "'openssl' not in ansible_facts.packages"
      delegate_to: localhost
      run_once: true
```
Fails with: This command has to be run under the root user.

richm:
> All CI failures are due to tasks in `tests/tasks/fixture_psks.yml`.
> 
> ```yaml
>     - name: List packages on the controller to see if OpenSSL is installed
>       package_facts:
>       delegate_to: localhost
>       run_once: true
> ```
> 
> Fails with: Could not detect a supported package manager from the following list: ['pkg', 'portage', 'apt', 'rpm'], or the required Python library is not installed. Check warnings for details.
> 
> ```yaml
>     - name: Ensure OpenSSL is installed on the controller
>       package:
>         name: openssl
>         state: present
>       when: "'openssl' not in ansible_facts.packages"
>       delegate_to: localhost
>       run_once: true
> ```
> 
> Fails with: This command has to be run under the root user.

Right.  Ignore those.  We're trying to figure out how best to install openssl on the controller hosts.

PR #63, opened 2022-08-24
Use the firewall role and the selinux role from the ha_cluster role

nhosoi:
- Introduce ha_cluster_manage_firewall to use the firewall role to manage the high-availability service and the fence-virt port. ha_cluster_manage_firewall is set to true, by default.
 - Introduce ha_cluster_manage_selinux to use the selinux role to manage the ports in the high-availability service. ha_cluster_manage_selinux is set to true, by default.
- Add the test check task tasks/check_firewall_selinux.yml for verify the ports status.
- Add meta/collection-requirements.yml.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,42 @@ An Ansible role for managing High Availability Clustering.
 
 boolean, default: `true`
 
-RHEL and CentOS only, enable repositories contaning needed packages
+RHEL and CentOS only, enable repositories containing needed packages
+
+#### `ha_cluster_manage_firewall`
+
+boolean, default: false
+
+Manage the `firewall high-availability service` as well as the `fence-virt port`.
+When `ha_cluster_manage_firewall` is `true`, the `firewall high-availability
+service` and `fence-virt port` are enabled.
+When `ha_cluster_manage_firewall` is `false`, the `ha_cluster role` does not
+manage the firewall.
+
+NOTE: `ha_cluster_manage_firewall` is limited to *adding* ports.
+It cannot be used for *removing* ports.
+If you want to remove ports, you will need to use the firewall system
+role directly.
+
+#### `ha_cluster_manage_selinux`
+
+boolean, default: false
+
+Manage the ports belonging to the `firewall high-availability service` using
+the selinux role.
+When `ha_cluster_manage_selinux` is `true`, the ports belonging to the
+`firewall high-availability service` are associated with the selinux port type
+`cluster_port_t`.
+When `ha_cluster_manage_selinux` is `false`, the `ha_cluster role` does not
+manage the selinux.
+
+NOTE: The firewall configuration is prerequisite for managing selinux. If the
+firewall is not installed, managing selinux policy is skipped.
+
+NOTE: `ha_cluster_manage_selinux` is limited to *adding* policy.
+It cannot be used for *removing* policy.
+If you want to remove policy, you will need to use the selinux system
+role directly.
 
 #### `ha_cluster_cluster_present`
 
@@ -58,7 +93,7 @@ automatically by the role, for example custom resource agents.
 
 It is possible to specify fence agents here as well. However,
 `ha_cluster_fence_agent_packages` is preferred for that, so that its default
-value is overriden.
+value is overridden.
 
 #### `ha_cluster_hacluster_password`
 
@@ -870,7 +905,7 @@ all:
 
 #### SBD watchdog and devices
 When using SBD, you may optionally configure watchdog and SBD devices for each
-node in inventory. Even though all SBD devices must be shared to and accesible
+node in inventory. Even though all SBD devices must be shared to and accessible
 from all nodes, each node may use different names for the devices. Watchdog may
 be different for each node as well. See also [SBD
 variables](#ha_cluster_sbd_enabled).
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -50,3 +50,11 @@ ha_cluster_constraints_location: []
 ha_cluster_constraints_colocation: []
 ha_cluster_constraints_order: []
 ha_cluster_constraints_ticket: []
+
+# If true, manage the high-availability service and the fence-virt port
+# using the firewall role.
+ha_cluster_manage_firewall: false
+
+# If true, manage the ports belonging to the high-availability service
+# and the fence-virt using the selinux role.
+ha_cluster_manage_selinux: false
--- a/tasks/firewall.yml
+++ b/tasks/firewall.yml
@@ -1,70 +1,18 @@
 # SPDX-License-Identifier: MIT
 ---
-- name: Get services status - detect firewall
-  service_facts:
-
-- name: Configure firewalld
-  block:
-    - name: Enable service 'high-availability' in firewalld
-      command: firewall-cmd --add-service high-availability
-      when: ansible_facts.services['firewalld.service'].state == 'running'
-      register: __ha_cluster_firewall_cmd
-      failed_when:
-        - __ha_cluster_firewall_cmd.rc != 0
-        - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
-    - name: Enable service 'high-availability' in firewalld permanent config
-      command: firewall-cmd --permanent --add-service high-availability
-      register: __ha_cluster_firewall_cmd
-      when: ansible_facts.services['firewalld.service'].state == 'running'
-      failed_when:
-        - __ha_cluster_firewall_cmd.rc != 0
-        - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
-    - name: Enable service 'high-availability' in firewalld offline config
-      command: firewall-offline-cmd --add-service high-availability
-      register: __ha_cluster_firewall_cmd
-      when: ansible_facts.services['firewalld.service'].state != 'running'
-      failed_when:
-        - __ha_cluster_firewall_cmd.rc != 0
-        - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
-    - name: Enable fence-virt port in firewalld - all options
-      when:
-        - ansible_architecture == 'x86_64'
-        - (
-            'fence-virt' in ha_cluster_fence_agent_packages
-            or
-            'fence-virt' in ha_cluster_extra_packages
-            or
-            'fence-agents-all' in ha_cluster_fence_agent_packages
-            or
-            'fence-agents-all' in ha_cluster_extra_packages
-          )
-      block:
-        - name: Enable fence-virt port in firewalld
-          command: firewall-cmd --add-port 1229/tcp
-          when: ansible_facts.services['firewalld.service'].state == 'running'
-          register: __ha_cluster_firewall_cmd
-          failed_when:
-            - __ha_cluster_firewall_cmd.rc != 0
-            - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
-        - name: Enable fence-virt port in firewalld permanent config
-          command: firewall-cmd --permanent --add-port 1229/tcp
-          when: ansible_facts.services['firewalld.service'].state == 'running'
-          register: __ha_cluster_firewall_cmd
-          failed_when:
-            - __ha_cluster_firewall_cmd.rc != 0
-            - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
-        - name: Enable fence-virt port in firewalld offline config
-          command: firewall-offline-cmd --add-port 1229/tcp
-          when: ansible_facts.services['firewalld.service'].state != 'running'
-          register: __ha_cluster_firewall_cmd
-          failed_when:
-            - __ha_cluster_firewall_cmd.rc != 0
-            - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
+- name: Ensure the service and the ports status with the firewall role
+  include_role:
+    name: fedora.linux_system_roles.firewall
+  vars:
+    __arch: "{{ ansible_facts['architecture'] }}"
+    __use_fence_fw_port: "{{ __arch == 'x86_64' and
+      ('fence-virt' in ha_cluster_fence_agent_packages
+       or 'fence-virt' in ha_cluster_extra_packages
+       or 'fence-agents-all' in ha_cluster_fence_agent_packages
+       or 'fence-agents-all' in ha_cluster_extra_packages) }}"
+    __fence_fw_port: "{{ [{'port': '1229/tcp', 'state': 'enabled'}]
+      if __use_fence_fw_port else [] }}"
+    __ha_fw_service: [{'service': 'high-availability', 'state': 'enabled' }]
+    firewall: "{{ __ha_fw_service + __fence_fw_port }}"
   when:
-    - '"firewalld.service" in ansible_facts.services'
+    - ha_cluster_manage_firewall | bool
--- a/tasks/selinux.yml
+++ b/tasks/selinux.yml
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Populate service facts
+  service_facts:
+
+- block:
+    - name: Set the fence-virt/fence-agents port to _ha_cluster_selinux
+      set_fact:
+        _ha_cluster_selinux:
+          - {'ports': '1229', 'proto': 'tcp', 'setype': 'cluster_port_t',
+             'state': 'present', 'local': true}
+      when:
+        - (
+            'fence-virt' in ha_cluster_fence_agent_packages
+            or
+            'fence-virt' in ha_cluster_extra_packages
+            or
+            'fence-agents-all' in ha_cluster_fence_agent_packages
+            or
+            'fence-agents-all' in ha_cluster_extra_packages
+          )
+
+    - name: Get associated selinux ports
+      shell: |-
+        set -euo pipefail
+        firewall-cmd --info-service=high-availability | \
+          egrep " +ports: +" | sed -e "s/ *ports: //"
+      register: __ports
+      changed_when: false
+
+    - name: Add the high-availability service ports to _ha_cluster_selinux
+      set_fact:
+        _ha_cluster_selinux: "{{ _ha_cluster_selinux | d([]) +
+            [{'ports': _pair[0], 'proto': _pair[1], 'setype': 'cluster_port_t',
+              'state': 'present', 'local': true}] }}"
+      vars:
+        _pair: "{{ item.split('/') }}"
+      loop: "{{ __ports.stdout.split(' ') }}"
+
+    - name: Ensure the service and the ports status with the selinux role
+      include_role:
+        name: fedora.linux_system_roles.selinux
+      vars:
+        selinux_ports: "{{ _ha_cluster_selinux }}"
+  when:
+    - '"firewalld.service" in ansible_facts.services'
+    - ansible_facts.services["firewalld.service"]["state"] == "running"
+    - ha_cluster_manage_selinux | bool
--- a/meta/collection-requirements.yml
+++ b/meta/collection-requirements.yml
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: MIT
+collections:
+  - fedora.linux_system_roles
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,6 +14,9 @@
     - name: Configure firewall
       include_tasks: firewall.yml
 
+    - name: Configure selinux
+      include_tasks: selinux.yml
+
     - name: Configure pcs / pcsd
       include_tasks: pcs-configure-pcs-pcsd.yml
 
--- a/tests/tasks/check_firewall_selinux.yml
+++ b/tests/tasks/check_firewall_selinux.yml
@@ -1,0 +1,38 @@
+# SPDX-License-Identifier: MIT
+---
+- block:
+    - name: Check firewall service status
+      command: firewall-cmd --list-services
+      register: _result
+      failed_when: "'high-availability' not in _result.stdout"
+      changed_when: false
+
+    - name: Check firewall port status
+      command: firewall-cmd --list-ports
+      register: _result
+      failed_when: "'1229/tcp' not in _result.stdout"
+      changed_when: false
+  when: ha_cluster_manage_firewall | bool
+
+- block:
+    - name: Get associated selinux ports
+      shell: |-
+        set -euo pipefail
+        firewall-cmd --info-service=high-availability | \
+          egrep " +ports: +" | sed -e "s/ *ports: //"
+      register: __ports
+      changed_when: false
+
+    - name: Check associated selinux ports
+      shell: |-
+        set -euo pipefail
+        sudo semanage port --list | grep cluster_port_t | \
+          grep "{{ _pair[0] }}" | grep "{{ _pair[1] }}"
+      vars:
+        _pair: "{{ item.split('/') }}"
+      loop: "{{ __ports.stdout.split(' ') }}"
+      changed_when: false
+  when:
+    - '"firewalld.service" in ansible_facts.services'
+    - ansible_facts.services["firewalld.service"]["state"] == "running"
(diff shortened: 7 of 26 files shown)


README.md
```diff
@@ -836,6 +836,12 @@ ha_cluster_constraints_ticket:
 You may take a look at
 [an example](#creating-a-cluster-with-resource-constraints).
 
+#### `ha_cluster_manage_firewall`
+
+boolean, default: `false`
+
+Manage the `firewall high-availability service` as well as the `fence-virt port`.
```
spetrosi, line 843: Now that firewall is not configured by default, README must also specify what users do when they want to manage firewall manually. I.e. enable the high-availability service and the 1229/tcp port.

tasks/firewall.yml
```diff
@@ -40,30 +20,12 @@
...
-  when:
-    - '"firewalld.service" in ansible_facts.services'
+    - name: Ensure the service and the ports status with the firewall role
+      include_role:
+        name: fedora.linux_system_roles.firewall
+      vars:
+        firewall: "{{ _ha_cluster_firewall }}"
+  vars:
+    _state_value: "{{ 'enabled' if ha_cluster_manage_firewall |
+                   bool | d(false) else 'disabled' }}"
```
spetrosi, line 31: Why do you need `d(false)` here if the ha_cluster_manage_firewall is set to `false` in defaults/main.yml?

README.md
```diff
@@ -836,6 +836,12 @@ ha_cluster_constraints_ticket:
 You may take a look at
 [an example](#creating-a-cluster-with-resource-constraints).
 
+#### `ha_cluster_manage_firewall`
```
spetrosi, line 839: In mssql, I use `mssql_firewall_configure` variable for this purpose. That's because historically, I was using the same format for `mssql_ha_configure`. Do we want to make this consistent across all roles?

README.md
richm: If firewall was previously configured by default, then we should make the default value for `ha_cluster_manage_firewall` to be `true`

nhosoi: Right. That's what I was worried. Otherwise `ha_cluster_manage_firewall: false` would wipe it out... I'm adding the code.

nhosoi:
BTW, here's the list of the `semanage port -l` result of the ports associated with this firewall role support. As listed, most of them are not managed for selinux. (Please note that 1229, 5404 and 5405 are defined in the policy.) Do we want to add the selinux support to the rest of the ports? (I'd think the answer is yes.) If so, what port type we should give? `netsupport_port_t`?
```
# semanage port -l | egrep "1229|2224|3121|5403|5404|5405|21064|9929"
netsupport_port_t              tcp      5404, 5405
netsupport_port_t              udp      5404, 5405
zented_port_t                  tcp      1229
zented_port_t                  udp      1229
```

richm:
> BTW, here's the list of the `semanage port -l` result of the ports associated with this firewall role support. As listed, most of them are not managed for selinux. (Please note that 1229, 5404 and 5405 are defined in the policy.) Do we want to add the selinux support to the rest of the ports? (I'd think the answer is yes.) If so, what port type we should give? `netsupport_port_t`?
> 
> ```
> # semanage port -l | egrep "1229|2224|3121|5403|5404|5405|21064|9929"
> netsupport_port_t              tcp      5404, 5405
> netsupport_port_t              udp      5404, 5405
> zented_port_t                  tcp      1229
> zented_port_t                  udp      1229
> ```

Not sure - @tomjelinek ?

README.md
nhosoi: Please take a look at https://github.com/linux-system-roles/ha_cluster/pull/63/commits/0ae86fb860df3ff09717d38e34e244deabad1080.

richm: I would prefer to have `ha_cluster_manage_firewall` default to `true`
and just have
```
if ha_cluster_manage_firewall is true:
    enable firewall ports
else:
    disable firewall ports
```
I'm not sure what is the problem that we're trying to solve by adding the more complex code.

nhosoi: If the default value is `true`, it always configures and enables the firewall, which is different from the current behavior. Is it ok?

richm: > If the default value is `true`, it always configures and enables the firewall, which is different from the current behavior. Is it ok?

I think so - I think that's what the author originally intended (@tomjelinek ?), but we didn't have a firewall role at that time.

nhosoi: Assuming it's ok to always configure firewall, I pushed a new commit. Thanks!

nhosoi:
This PR will be updated once this selinux [pr/122](https://github.com/linux-system-roles/selinux/pull/122) is completed and merged.

nhosoi:
Notes: 
- This pr depends upon https://github.com/linux-system-roles/selinux/pull/122
- The setype of the ports in the `high-availability service` is set to `netsupport_port_t`.

defaults/main.yml
```diff
@@ -48,3 +48,11 @@ ha_cluster_constraints_location: []
...
 ha_cluster_constraints_order: []
 ha_cluster_constraints_ticket: []
+
+# If true, manage the high-availability service and the fence-virt port
+# using the firewall role.
+ha_cluster_manage_firewall: true
+
+# If true, manage the ports belonging to the high-availability service
+# and the fence-virt using the selinux role.
+ha_cluster_manage_selinux: true
```
tomjelinek, line 58: Do we want this variable to be documented in README.md? If not, can we consider moving it to vars/main.yml?

README.md
tomjelinek: I would put this variable description to the top, perhaps right bellow `ha_cluster_enable_repos`. That's where system-like settings are described. Here, at the bottom, cluster configuration is documented.

nhosoi: Thanks a lot, @tomjelinek! I've pushed another commit to fix it. (I will or please do squash before merging this pr.)

defaults/main.yml
nhosoi: Thanks a lot, @tomjelinek! I've pushed another commit to fix it. (I will or please do squash before merging this pr.)

Please feel free to improve the sentences.

README.md
nhosoi: Thank you for your input, @tomjelinek. Glad to hear it does not break our customers' needs.

nhosoi:
Thanks so much for reviewing this pr, @tomjelinek! I've squashed the multiple commits into one.

Actually, due to the BaseOS CI status, we have not run the CI tests yet.
I'm going to launch it manually. This pr requires new enhancements in the BaseOS CI. Unless they are provisioned in the production system, the tests for this pr would fail. I'd think we should wait to merge this pr until the tests on the BaseOS CI env successfully pass.

nhosoi:
[citest]

richm:
You will probably need to rebase now that https://github.com/linux-system-roles/ha_cluster/pull/64 is merged

nhosoi:
> You will probably need to rebase now that #64 is merged

Thank you, @richm. I will.

Before doing that, I'd like to discuss about having two variables `ha_cluster_manage_firewall` and `ha_cluster_manage_selinux`, which is shared with other roles, such as metrics and vpn (so far :).

The issue is when the ports are managed by the firewall service (in ha_cluster case `high-availability`), we read the firewall service xml file to get the port numbers. If `ha_cluster_manage_firewall` is false, it may no exist unless the firewalld is already installed on the managed host prior to running the `ha_cluster` role. In that case, we have to give up the selinux configuration. Do you think having the two variables is reasonable? Or we'd better have just one to cover both? Can we think of a use case configuring just one of firewalld and selinux and not the other???

tasks/selinux.yml
```diff
@@ -0,0 +1,57 @@
...
+            'fence-virt' in ha_cluster_fence_agent_packages
+            or
+            'fence-virt' in ha_cluster_extra_packages
+            or
+            'fence-agents-all' in ha_cluster_fence_agent_packages
+            or
+            'fence-agents-all' in ha_cluster_extra_packages
+          )
+
+    - name: Get the high-availability tcp service ports
```
richm, line 20: you can use `firewall-cmd --service=high-availability --get-ports` to get the ports - here is the output:
```
>sudo firewall-cmd --info-service=high-availability
high-availability
  ports: 2224/tcp 3121/tcp 5403/tcp 5404/udp 5405-5412/udp 9929/tcp 9929/udp 21064/tcp
  protocols: 
  source-ports: 
  modules: 
  destination: 
  includes: 
  helpers: 
```

richm:
> > You will probably need to rebase now that #64 is merged
> 
> Thank you, @richm. I will.
> 
> Before doing that, I'd like to discuss about having two variables `ha_cluster_manage_firewall` and `ha_cluster_manage_selinux`, which is shared with other roles, such as metrics and vpn (so far :).
> 
> The issue is when the ports are managed by the firewall service (in ha_cluster case `high-availability`), we read the firewall service xml file to get the port numbers. If `ha_cluster_manage_firewall` is false, it may no exist unless the firewalld is already installed on the managed host prior to running the `ha_cluster` role.

Or - the ha_cluster role will have to use the `firewall` role first, if enabled.

> In that case, we have to give up the selinux configuration. Do you think having the two variables is reasonable? Or we'd better have just one to cover both? Can we think of a use case configuring just one of firewalld and selinux and not the other???

I think we should have both - there may be a case for the user wanting `ha_cluster_manage_firewall` without `ha_cluster_manage_selinux`.  However, as you point out, there is no way to dynamically determine the `high-availability` service ports without firewall being installed.  In that case, if the user specifies `ha_cluster_manage_selinux: true`, we should always use `firewall-cmd --service=high-availability --get-ports` to get the ports, and the role will fail if that command fails.

nhosoi:
> I think we should have both - there may be a case for the user wanting `ha_cluster_manage_firewall` without `ha_cluster_manage_selinux`. 

Ok!

> However, as you point out, there is no way to dynamically determine the `high-availability` service ports without firewall being installed. In that case, if the user specifies `ha_cluster_manage_selinux: true`, we should always use `firewall-cmd --service=high-availability --get-ports` to get the ports, and the role will fail if that command fails.

You mean, it's ok to fail as long as we tell the customer to specify `ha_cluster_manage_firewall: true` (maybe in the error message?) if they want to configure `selinux`?

richm:
> > I think we should have both - there may be a case for the user wanting `ha_cluster_manage_firewall` without `ha_cluster_manage_selinux`.
> 
> Ok!
> 
> > However, as you point out, there is no way to dynamically determine the `high-availability` service ports without firewall being installed. In that case, if the user specifies `ha_cluster_manage_selinux: true`, we should always use `firewall-cmd --service=high-availability --get-ports` to get the ports, and the role will fail if that command fails.
> 
> You mean, it's ok to fail as long as we tell the customer to specify `ha_cluster_manage_firewall: true` (maybe in the error message?) if they want to configure `selinux`?

Yes, in the README and in the error message.

tasks/selinux.yml
richm: In fact, you can eliminate using `set_fact` altogether:
```yaml
    - name: get firewall ports
      command: firewall-cmd --info-service=high-availability
      register: __out
    - name: construct selinux args
      debug:
        msg: selinux_ports {{ selinux_ports | to_nice_json }}
      vars:
        _state_value: enabled
        __pat: '(?sm)^.* ports: ([^\n]+)\n.*$'
        __port_list: "{{ __out.stdout | regex_replace(__pat, '\\1') | split |
          map('split', '/') }}"
        selinux_ports: |
          {% set selinux_ports = [] %}
          {% for port_proto in __port_list %}
          {%   set item = {'ports': port_proto[0], 'proto': port_proto[1],
                           'setype': 'cluster_port_t',
                           'state': _state_value, 'local': true} %}
          {%   set _ = selinux_ports.append(item) %}
          {% endfor %}
          {{ selinux_ports }}
```

nhosoi:
Sorry... I must have been sleeping... In the current implementation, there's no way not to install `firewalld`. If `ha_cluster_manage_firewall: true`, it enables the high-availability ports. If `false`, it disables them. (and `true`, by default.) So, there's no worry about not having `firewall-cmd`. Hope it's ok...

richm:
> Sorry... I must have been sleeping... In the current implementation, there's no way not to install `firewalld`. If `ha_cluster_manage_firewall: true`, it enables the high-availability ports. If `false`, it disables them. (and `true`, by default.) So, there's no worry about not having `firewall-cmd`. Hope it's ok...

Yes, that's ok.

richm:
[citest]

nhosoi:
[citest]

README.md
```diff
@@ -32,29 +32,38 @@ RHEL and CentOS only, enable repositories containing needed packages
...
-When `ha_cluster_manage_firewall` is `false`, the `firewall high-availability
-service` and `fence-virt port` are disabled.
-Please note that configuring firewalld is ensured regardless of the
-`ha_cluster_manage_firewall` value.
+When `ha_cluster_manage_firewall` is `false`, the `ha_cluster role` does not
+manage the firewall.
+
+NOTE: `ha_cluster_manage_firewall` is limited to *adding* ports.
+It cannot be used for *removing* ports.
+If you want to remove ports, you will need to use the firewall selinux system
```
richm, line 45: "use the firewall and selinux system"

tasks/firewall.yml
```diff
@@ -27,6 +27,5 @@
         name: fedora.linux_system_roles.firewall
       vars:
         firewall: "{{ _ha_cluster_firewall }}"
-  vars:
-    _state_value: "{{ ha_cluster_manage_firewall |
-                      ternary('enabled', 'disabled') }}"
+  when:
```
richm, line 30: would prefer to do this without `set_fact`:
```yaml
---
- name: Ensure the service and the ports status with the firewall role
  include_role:
    name: fedora.linux_system_roles.firewall
  vars:
    __use_fence_fw_port: "{{ __arch == 'x86_64' and
      ('fence-virt' in ha_cluster_fence_agent_packages
       or 'fence-virt' in ha_cluster_extra_packages
       or 'fence-agents-all' in ha_cluster_fence_agent_packages
       or 'fence-agents-all' in ha_cluster_extra_packages) }}"
    __fence_fw_port: "{{ [{'port': '1229/tcp', 'state': 'enabled'}]
      if __use_fence_fw_port else [] }}"
    __ha_fw_service: [{'service': 'high-availability', 'state': 'enabled' }]
    firewall: "{{ __ha_fw_service + __fence_fw_port }}"
when:
    - ha_cluster_manage_firewall | bool
```

tasks/selinux.yml
```diff
@@ -1,11 +1,14 @@
...
+- name: Populate service facts
+  service_facts:
+
 - block:
     - name: Set the fence-virt/fence-agents port to _ha_cluster_selinux
       set_fact:
         _ha_cluster_selinux:
           - {'ports': '1229', 'proto': 'tcp', 'setype': 'cluster_port_t',
-             'state': '{{ _state_value }}', 'local': 'true'}
+             'state': 'present', 'local': 'true'}
```
richm, line 11: I think this should be `'local': true` - that is, the boolean value `true` instead of the string value `'true'` - also below

tasks/selinux.yml
```diff
@@ -39,6 +42,6 @@
         name: fedora.linux_system_roles.selinux
       vars:
         selinux_ports: "{{ _ha_cluster_selinux }}"
-  vars:
-    _state_value: "{{ ha_cluster_manage_selinux |
-                      ternary('present', 'absent') }}"
+  when:
+    - '"firewalld.service" in ansible_facts.services'
+    - ha_cluster_manage_firewall | bool
```
richm, line 47: `ha_cluster_manage_selinux | bool`

tasks/firewall.yml
nhosoi: I was not sure what's the best way to get the `__arch` value. So, I set it to `"{{ ansible_facts['architecture'] }}"`. Is it ok?
```
  6   vars:
  7     __arch: "{{ ansible_facts['architecture'] }}"
  8     __use_fence_fw_port: "{{ __arch == 'x86_64' and
```

README.md
```diff
@@ -42,7 +42,7 @@ manage the firewall.
 
 NOTE: `ha_cluster_manage_firewall` is limited to *adding* ports.
 It cannot be used for *removing* ports.
-If you want to remove ports, you will need to use the firewall selinux system
+If you want to remove ports, you will need to use the firewall system
 roles directly.
```
richm, line 46:
```suggestion
role directly.
```

README.md
```diff
@@ -62,7 +62,7 @@ firewall is not installed, managing selinux policy is skipped.
 
 NOTE: `ha_cluster_manage_selinux` is limited to *adding* policy.
 It cannot be used for *removing* policy.
-If you want to remove policy, you will need to use the selinux selinux system
+If you want to remove policy, you will need to use the selinux system
 roles directly.
```
richm, line 66:
```suggestion
role directly.
```

tasks/firewall.yml
richm: > I was not sure what's the best way to get the `__arch` value. So, I set it to `"{{ ansible_facts['architecture'] }}"`. Is it ok?
> 
> ```
>   6   vars:
>   7     __arch: "{{ ansible_facts['architecture'] }}"
>   8     __use_fence_fw_port: "{{ __arch == 'x86_64' and
> ```

Yes.

nhosoi:
[citest]

richm (approved):
May take several hours for CI tests to run . . .

ukulekek:
[citest]

ukulekek:
Sorry, needed to rerun CI, we had an outage at that time

tasks/selinux.yml
```diff
@@ -0,0 +1,47 @@
...
+        _pair: "{{ item.split('/') }}"
+      loop: "{{ __ports.stdout.split(' ') }}"
+
+    - name: Ensure the service and the ports status with the selinux role
+      include_role:
+        name: fedora.linux_system_roles.selinux
+      vars:
+        selinux_ports: "{{ _ha_cluster_selinux }}"
+  when:
+    - '"firewalld.service" in ansible_facts.services'
```
richm, line 46: hmm - seems this condition is not sufficient - not only does the firewalld.service have to be present, it must also be running - here is the state in the test: https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/lsr-citool_ha_cluster-63-ed61ee7_CentOS-Stream-8_20220923-130045/artifacts/tests_cib_constraints_create-FAILED.log
```
            "firewalld.service": {
                "name": "firewalld.service",
                "source": "systemd",
                "state": "inactive",
                "status": "disabled"
            },
```
here is what a running service looks like:
```
            "crond.service": {
                "name": "crond.service",
                "source": "systemd",
                "state": "running",
                "status": "enabled"
            },
```
so maybe a condition like
```suggestion
    - '"firewalld.service" in ansible_facts.services
    - ansible_facts.services["firewalld.service"]["state"] == "running"
```

nhosoi: Thank you, @richm. I will fix it as you suggested.
I'm wondering why I could not capture this failure in this local test... (;_;)
```
tox -e qemu-ansible-core-2.13 -- --image-name rhel-9 --log-level info tests/tests_cib_constraints_create.yml
```

richm: > Thank you, @richm. I will fix it as you suggested. I'm wondering why I could not capture this failure in this local test... (;_;)
> 
> ```
> tox -e qemu-ansible-core-2.13 -- --image-name rhel-9 --log-level info tests/tests_cib_constraints_create.yml
> ```

It must have something to do with how the images are built.  I'm guessing the qcow2 images we use from the composes are built with firewalld active by default, but the openstack images used by baseos ci do not have firewalld active by default?

nhosoi: Ok. I removed firewalld before running the test and made sure the test passed. (no worry, I do not plan to include it in the CI. :)
```
    - name: Remove firewall package
      package:
        name: firewalld
        state: absent
```

nhosoi:
[citest]

richm:
@nhosoi looks good - ready to merge?

nhosoi:
> @nhosoi looks good - ready to merge?

Yes, please merge. Thanks a lot for your review comments, @richm, @tomjelinek, and @spetrosi.

tasks/firewall.yml
```diff
@@ -1,70 +1,18 @@
...
-          command: firewall-offline-cmd --add-port 1229/tcp
-          when: ansible_facts.services['firewalld.service'].state != 'running'
-          register: __ha_cluster_firewall_cmd
-          failed_when:
-            - __ha_cluster_firewall_cmd.rc != 0
-            - __ha_cluster_firewall_cmd.rc != 11  # already enabled
-
+- name: Ensure the service and the ports status with the firewall role
+  include_role:
+    name: fedora.linux_system_roles.firewall
```
tomjelinek, line 5: This causes the role to fail on my VMs as the firewall role cannot be found. It seems like ansible collections need to be installed for this to work. Is the `rhel-system-roles` packages going to depend on ansible collections? Should this be `linux-system-roles.firewall` instead, so that it uses the firewall role from the `rhel-system-roles` package?

tasks/selinux.yml
```diff
@@ -0,0 +1,48 @@
...
+        _ha_cluster_selinux: "{{ _ha_cluster_selinux | d([]) +
+            [{'ports': _pair[0], 'proto': _pair[1], 'setype': 'cluster_port_t',
+              'state': 'present', 'local': true}] }}"
+      vars:
+        _pair: "{{ item.split('/') }}"
+      loop: "{{ __ports.stdout.split(' ') }}"
+
+    - name: Ensure the service and the ports status with the selinux role
+      include_role:
+        name: fedora.linux_system_roles.selinux
```
tomjelinek, line 42: See comment above about firewall role. Should this be `linux-system-roles.selinux` instead?

tasks/firewall.yml
richm: > This causes the role to fail on my VMs

How are you running your VMs?  The `tox-lsr` tox plugin using `-e qemu[-ansible-core-2.x]` supports installing collection dependencies for the test.
We are following https://linux-system-roles.github.io/documentation/role-requirements.html
And if you use https://github.com/linux-system-roles/tox-lsr#qemu-testing e.g.
`tox -e qemu-ansible-core-2.13 -- --image-name centos-9 tests/tests_cluster_basic_new_psks.yml`
then tox will automatically install the dependencies from `meta/collection-requirements.yml` in your test environment (i.e. not in ~/.ansible/collections and "pollute" your local workstation).

> as the firewall role cannot be found. It seems like ansible collections need to be installed for this to work.

Yes.

> Is the `rhel-system-roles` packages going to depend on ansible collections?

For upstream (users installing the ha_cluster role from Galaxy), users will need to follow directions like this: https://github.com/linux-system-roles/storage/#requirements

@nhosoi The ha_cluster README needs to be updated with a `Requirements` section

Users using the ha_cluster role as part of the fedora.linux_system_roles collection will not need to do this, nor users using the role from the Fedora RPM package.

For downstream, the dependencies will be automatically handled, so the user will need to do nothing.

> Should this be `linux-system-roles.firewall` instead, so that it uses the firewall role from the `rhel-system-roles` package?

No, see above.

tasks/selinux.yml
richm: > See comment above about firewall role. Should this be `linux-system-roles.selinux` instead?

No, see above.

tasks/firewall.yml
tomjelinek: Ok, thanks for explaining. I don't use tox. I'm actually developing and testing the role on VMs which I use for developing pcs as well. Once I installed the collection on the VMs, everything started working just fine.

nhosoi: Thank you for the confirmation, @tomjelinek!
I'm updating the README.md with the new requirements and the examples as suggested by @richm and @spetrosi. (stay tuned...)

nhosoi: @tomjelinek, @richm, @spetrosi, could you please take a look at https://github.com/linux-system-roles/ha_cluster/pull/69?
Thanks!

PR #68, opened 2022-09-22
Add support for quorum device

tomjelinek:
With this update, the role is capable of configuring a quorum device. The device is a daemon meant to run on a host outside of a cluster. It provides an additional quorum vote and can help dealing with network splits.

As seen in README.md, the role is run once to configure a quorum device host, and then with a different playbook to configure a cluster. Is this ok? Would it be better to have a separate role for configuring a quorum device host instead?


richm:
> With this update, the role is capable of configuring a quorum device. The device is a daemon meant to run on a host outside of a cluster. It provides an additional quorum vote and can help dealing with network splits.
> 
> As seen in README.md, the role is run once to configure a quorum device host, and then with a different playbook to configure a cluster. Is this ok? Would it be better to have a separate role for configuring a quorum device host instead?

I think a single role is ok.  In this case, the user would define two separate sets of host on which to run the ha_cluster role: one host which is the quorum device host, and one host or group of hosts on which to configure for the HA cluster.  Something like this e.g. in a YAML inventory file:
```yaml
all:
  hosts:
.... the hosts ....
  children:
    ha_cluster_quorum_device_host:
      hosts:
        the_name_of_the_host_from_hosts:
      vars:
        ha_cluster_qnetd:
          present: boolean
          start_on_boot: boolean
          regenerate_keys: boolean
    ha_cluster_hosts:
      hosts:
        host1:
        host2:
      vars:
        ha_cluster_hacluster_password: ....
        ha_cluster_cluster_name: ....
        ha_cluster_other_parameters: to apply to all hosts in the ha_cluster_hosts group
```
Then the playbook would call the role like this - the ha_cluster_quorum_device_host settings would be applied first, then ha_cluster_hosts:
```yaml
# set up HA clustering
- hosts: ha_cluster_quorum_device_host ha_cluster_hosts
  order
  roles:
```
If for some reason this doesn't preserve or guarantee the order, the user may have to structure their inventory correctly, or otherwise use one of the ordering strategies listed at https://docs.ansible.com/ansible/latest/user_guide/playbooks_strategies.html#ordering-execution-based-on-inventory

NOTE: `ha_cluster_hosts` can contain the host used in `ha_cluster_quorum_device_host` - it will apply the quorum device settings to the host first, then the "regular" ha_cluster settings.

tomjelinek:
[citest]

tomjelinek:
[citest]

tomjelinek:
[citest]

tomjelinek:
[citest]

tomjelinek:
[citest]

tomjelinek:
[citest]

spetrosi (approved):
lgtm

PR #69, opened 2022-09-26
Update README for "use other roles from ha_cluster role".

nhosoi:
- Add requirements
- Add an example

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,23 @@ An Ansible role for managing High Availability Clustering.
   * stonith and resources
   * resource constraints
 
+## Requirements
+
+The role requires the `firewall` role and the `selinux` role from the
+`fedora.linux_system_roles` collection, if `ha_cluster_manage_firewall`
+and `ha_cluster_manage_selinux` is set to true, respectively.
+(Please see also [`ha_cluster_manage_firewall`](#ha_cluster_manage_firewall)
+ and [`ha_cluster_manage_selinux`](#ha_cluster_manage_selinux))
+
+If the `ha_cluster` is a role from the `fedora.linux_system_roles`
+collection or from the Fedora RPM package, the requirement is already
+satisfied.
+
+Otherwise, please run the following command line to install the collection.
+```
+ansible-galaxy collection install -r meta/collection-requirements.yml
+```
+
 ## Role Variables
 
 ### Defined in `defaults/main.yml`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,6 +20,23 @@ An Ansible role for managing High Availability Clustering.
...
+(Please see also [`ha_cluster_manage_firewall`](#ha_cluster_manage_firewall)
+ and [`ha_cluster_manage_selinux`](#ha_cluster_manage_selinux))
+
+If the `ha_cluster` is a role from the `fedora.linux_system_roles`
+collection or from the Fedora RPM package, the requirement is already
+satisfied.
+
+Otherwise, please run the following command line to install the collection.
+```
+ansible-galaxy collection install -r meta/requirements.yml
```
richm, line 37: collection-requirements.yml

richm (approved):

README.md
```diff
@@ -1291,6 +1313,17 @@ They are not guides or best practices for configuring a cluster.
     - linux-system-roles.ha_cluster
 ```
 
+### Configuring firewall and selinux using each role
```
spetrosi, line 1316: I think that instead of having this example, the role must set `ha_cluster_manage_firewall: true and `ha_cluster_manage_selinux: true` in all other examples alongside required variables. Because now when these vars are set to `false` by default, other examples won't work properly. What do you think @nhosoi ?

nhosoi: Ah, I see... Thank you for the clarification, @spetrosi. It should be better as examples could be used in the copied and pasted manner... I guess we should even emphasize it more with stronger recommendation to set them `true`.

@spetrosi, are you willing to modify the README file or shall I work on it?

spetrosi: I guess that is required in all roles where you `_manage_firewall` and `_manage_selinux` are added. I would prefer you to do this because you added them in the first place, but I can too if you have a lot on your plate.

nhosoi: @spetrosi, I've submitted https://github.com/linux-system-roles/ha_cluster/pull/70 for the examples.
I did not add the manage firewall/selinux variables in each example in README.md, but did so in the playbooks in the `examples` directory. And I'm trying to convince customers to set the variables to true when they create their playbook.

Could you please review it and share your thoughts? (Needless to say :), please feel free to rewriting the sentences/examples in README. Thanks!

PR #71, opened 2022-10-07
Use the certificate role to create the cert and the key

nhosoi:
- Introduce a variable ha_cluster_pcsd_certificates to set the certificate_requests.
- Add a test test script tests/tests_cluster_basic_custom_pcsd_tls_cert_role.yml

I borrowed the mssql README in https://github.com/linux-system-roles/mssql/pull/125 written by @spetrosi. Thanks, Sergei!

@spetrosi, @tomjelinek, this is another piece of the effort to use other roles in the ha_cluster role. If you could review it, I'd greatly appreciate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ An Ansible role for managing High Availability Clustering.
 The role requires the `firewall` role and the `selinux` role from the
 `fedora.linux_system_roles` collection, if `ha_cluster_manage_firewall`
 and `ha_cluster_manage_selinux` is set to true, respectively.
-(Please see also [`ha_cluster_manage_firewall`](#ha_cluster_manage_firewall)
- and [`ha_cluster_manage_selinux`](#ha_cluster_manage_selinux))
+Please see also [`ha_cluster_manage_firewall`](#ha_cluster_manage_firewall)
+ and [`ha_cluster_manage_selinux`](#ha_cluster_manage_selinux).
 
 If the `ha_cluster` is a role from the `fedora.linux_system_roles`
 collection or from the Fedora RPM package, the requirement is already
@@ -195,6 +195,32 @@ https://docs.ansible.com/ansible/latest/user_guide/vault.html for details.
 If these variables are set, `ha_cluster_regenerate_keys` is ignored for this
 certificate - key pair.
 
+#### `ha_cluster_pcsd_certificates`
+
+If there is no pcsd private key and certificate, there are two ways to create them.
+
+One way is by setting `ha_cluster_pcsd_certificates` variable.
+Another way is by setting none of
+[`ha_cluster_pcsd_public_key_src` and `ha_cluster_pcsd_private_key_src`](#ha_cluster_pcsd_public_key_src-ha_cluster_pcsd_private_key_src) and `ha_cluster_pcsd_certificates`.
+
+If `ha_cluster_pcsd_certificates` is provided, the `certificate` role is internally
+used and it creates the private key and certificate for pcsd as defined.
+If none of the variables are provided, the `ha_cluster` role will create pcsd
+certificates via pcsd itself. 
+
+The value of `ha_cluster_pcsd_certificates` is set to the variable `certificate_requests`
+in the `certificate` role.
+For more information, see the `certificate_requests` section in the `certificate`
+role documentation.
+
+The default value is `[]`.
+
+Note: when you set this variable, you must not set `ha_cluster_pcsd_public_key_src`
+and `ha_cluster_pcsd_private_key_src` variables.
+
+If this variable is set, `ha_cluster_regenerate_keys` is ignored for this
+certificate - key pair.
+
 #### `ha_cluster_regenerate_keys`
 
 boolean, default: `false`
@@ -207,6 +233,7 @@ See also:
 [`ha_cluster_fence_virt_key_src`](#ha_cluster_fence_virt_key_src),
 [`ha_cluster_pcsd_public_key_src`](#ha_cluster_pcsd_public_key_src-ha_cluster_pcsd_private_key_src),
 [`ha_cluster_pcsd_private_key_src`](#ha_cluster_pcsd_public_key_src-ha_cluster_pcsd_private_key_src)
+[`ha_cluster_pcsd_certificates`](#ha_cluster_pcsd_certificates)
 
 #### `ha_cluster_pcs_permission_list`
 
@@ -1047,6 +1074,22 @@ to `true` in your playbooks using the `ha_cluster` role.
     ha_cluster_manage_firewall: true
     ha_cluster_manage_selinux: true
 
+  roles:
+    - linux-system-roles.ha_cluster
+```
+
+### Creating pcsd certificate and private key files using the `certificate` role
+
+This example creates self-signed pcsd certificate and private key files
+in /var/lib/pcsd with the file name FILENAME.crt and FILENAME.key, respectively.
+
+```yaml
+- hosts: node1 node2
+  vars:
+    ha_cluster_pcsd_certificates:
+      - name: FILENAME
+        common_name: "{{ ansible_hostname }}"
+        ca: self-sign
   roles:
     - linux-system-roles.ha_cluster
 ```
--- a/tasks/pcs-configure-pcs-pcsd.yml
+++ b/tasks/pcs-configure-pcs-pcsd.yml
@@ -19,6 +19,40 @@
   when:
     - ha_cluster_regenerate_keys
 
+- name: Create pcsd certificate and private_key files
+  when: ha_cluster_pcsd_certificates | length > 0
+  block:
+    - name: Get the stat of /var/lib/pcsd
+      stat:
+        path: /var/lib/pcsd
+      register: __pcsd_stat
+
+    - name: Allow certmonger to write into pcsd's certificate directory
+      file:
+        path: /var/lib/pcsd
+        state: directory
+        setype: cert_t
+        mode: "{{ __pcsd_stat.stat.mode }}"
+
+    - name: >-
+       Ensure the name of ha_cluster_pcsd_certificates is /var/lib/pcsd/pcsd;
+       Create certificates using the certificate role
+      include_role:
+        name: fedora.linux_system_roles.certificate
+      vars:
+        certificate_requests: |
+          {% for _cert in ha_cluster_pcsd_certificates %}
+          {%   set _ = _cert.__setitem__('name', '/var/lib/pcsd/pcsd') %}
+          {% endfor %}
+          {{ ha_cluster_pcsd_certificates }}
+  always:
+    - name: Set pcsd's certificate directory back to cluster_var_lib_t
+      file:
+        path: /var/lib/pcsd
+        state: directory
+        setype: cluster_var_lib_t
+        mode: "{{ __pcsd_stat.stat.mode }}"
+
 - name: Distribute pcsd TLS certificate and key
   block:
     - name: Distribute pcsd TLS private key
@@ -36,6 +70,7 @@
         group: root
         mode: 0600
   when:
+    - ha_cluster_pcsd_certificates | length == 0
     - ha_cluster_pcsd_public_key_src is string
     - ha_cluster_pcsd_public_key_src | length > 0
     - ha_cluster_pcsd_private_key_src is string
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -21,6 +21,7 @@ ha_cluster_pacemaker_key_src: null
 ha_cluster_fence_virt_key_src: null
 ha_cluster_pcsd_public_key_src: null
 ha_cluster_pcsd_private_key_src: null
+ha_cluster_pcsd_certificates: []
 
 ha_cluster_cluster_name: my-cluster
 
--- a/examples/create-pcsd-cert.yml
+++ b/examples/create-pcsd-cert.yml
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: MIT
+---
+- name: >-
+    Example ha_cluster role invocation with the pcsd cert
+    and key generation by the certificate role
+  hosts: all
+  vars:
+    ha_cluster_pcsd_certificates:
+      - name: pcsd_cert
+        common_name: "{{ ansible_hostname }}"
+        ca: self-sign
+  roles:
+    - linux-system-roles.ha_cluster
--- a/tasks/check-and-prepare-role-variables.yml
+++ b/tasks/check-and-prepare-role-variables.yml
@@ -87,3 +87,17 @@
     - ha_cluster_quorum.options | d([])
       | selectattr('name', 'match', '^auto_tie_breaker$')
       | map(attribute='value') | select('in', ['0', 0]) | list | length > 0
+
+- name: >-
+    Fail if ha_cluster_pcsd_public_key_src and ha_cluster_pcsd_private_key_src
+    are set along with ha_cluster_pcsd_certificates
+  fail:
+    msg: >-
+      Cannot set ha_cluster_pcsd_public_key_src and
+      ha_cluster_pcsd_private_key_src along with ha_cluster_pcsd_certificates.
+      Remove ha_cluster_pcsd_public_key_src and ha_cluster_pcsd_private_key_src
+      or ha_cluster_pcsd_certificates.
+  when:
+    - ha_cluster_pcsd_public_key_src is not none
+    - ha_cluster_pcsd_private_key_src is not none
+    - ha_cluster_pcsd_certificates | d([])
--- a/tests/tests_cluster_basic_custom_pcsd_tls_cert_role.yml
+++ b/tests/tests_cluster_basic_custom_pcsd_tls_cert_role.yml
@@ -1,0 +1,50 @@
+# SPDX-License-Identifier: MIT
+---
+- name: >-
+    Minimal cluster configuration, custom pcsd TLS certificate -
+    using the certificate role
+  hosts: all
+  vars_files: vars/main.yml
+  vars:
+    ha_cluster_cluster_name: test-cluster
+    __cert_name: pcsd_cert
+
+  tasks:
+    - block:
+        - name: Set up test environment
+          include_role:
+            name: linux-system-roles.ha_cluster
+            tasks_from: test_setup.yml
+
+        - name: Run HA Cluster role
+          include_role:
+            name: linux-system-roles.ha_cluster
+            public: true
+          vars:
+            ha_cluster_pcsd_certificates:
+              - name: "{{ __cert_name }}"
+                common_name: "{{ ansible_hostname }}"
+                ca: self-sign
+
+        - name: Stat pcsd TLS certificate
+          stat:
+            path: /var/lib/pcsd/pcsd.crt
+            checksum_algorithm: sha1
+          register: stat_pcsd_cert
+
+        - name: Stat pcsd TLS key
+          stat:
+            path: /var/lib/pcsd/pcsd.key
+            checksum_algorithm: sha1
+          register: stat_pcsd_key
+
+        - name: Check pre-shared keys and TLS certificates are present
+          assert:
+            that:
+              - stat_pcsd_cert.stat.exists
+              - stat_pcsd_key.stat.exists
+
+        - name: Check firewall and selinux state
+          include_tasks: tasks/check_firewall_selinux.yml
+
+      tags: tests::verify


nhosoi:
[citest]

nhosoi:
[citest]

tasks/pcs-configure-pcs-pcsd.yml
```diff
@@ -19,6 +19,33 @@
   when:
     - ha_cluster_regenerate_keys
 
+- name: Create certificate and private_key files
+  when: ha_cluster_pcsd_certificates | length > 0
+  block:
+    - name: Allow certmonger to write into pcsd's certificate directory
+      file:
```
nhosoi, line 26: Hi @tomjelinek,
First, I created the cert and the key in the default location `/etc/pki/tls/{certs,private}` and moved the files to `/var/lib/pcsd` [0]. Then, I thought it'd be simpler for the certificate role to place them in `/var/lib/pcsd` directly. But we get this error `risky-file-permissions File permissions unset or incorrect` by the change. Without giving the permission to the certificate role, writing to the directory fails.

Should I go back to the original plan [0]? Or could there be a way to tell the github action to allow it? Thanks!

nhosoi:
[citest]

tasks/pcs-configure-pcs-pcsd.yml
spetrosi: You must set the required `mode` with the `file` module

nhosoi: > You must set the required `mode` with the `file` module to fix this

Thank you, @spetrosi. Hmmm, my intention is assuming `/var/lib/pcsd` exists and change only the setype with other attributes unchanged using this feature...
```
If mode is not specified and the destination filesystem object does exist, the mode of the existing filesystem object will be used.
```
But if it solves the issue, let me play around it by getting the mode and resetting it.
Thanks!

nhosoi:
[citest]

nhosoi:
[citest]

spetrosi (approved):
lgtm

README.md
```diff
@@ -194,6 +194,26 @@ https://docs.ansible.com/ansible/latest/user_guide/vault.html for details.
...
+Use this variable to generate certificate and private key for TLS encryption
+using the `fedora.linux_system_roles.certificate`.
+
+The value of `ha_cluster_pcsd_certificates` is set to the variable `certificate_requests`
+in the `certificate` role.
+For more information, see the `certificate_requests` section in the `certificate`
+role documentation.
+
+When you set this variable, you must not set `ha_cluster_pcsd_public_key_src` and
+`ha_cluster_pcsd_private_key_src` variables.
```
tomjelinek, line 212: It would be nice if you could add a check for this into `check-and-prepare-role-variables.yml`.

README.md
```diff
@@ -194,6 +194,26 @@ https://docs.ansible.com/ansible/latest/user_guide/vault.html for details.
...
+
+If this variable is set, `ha_cluster_regenerate_keys` is ignored for this
+certificate - key pair.
+
```
tomjelinek, line 216: Do we want an example in `README.md` and an example playbook to show how the `ha_cluster_pcsd_certificates` variable is used?

tasks/pcs-configure-pcs-pcsd.yml
```diff
@@ -19,6 +19,40 @@
   when:
     - ha_cluster_regenerate_keys
 
+- name: Create certificate and private_key files
```
tomjelinek, line 22:
```suggestion
- name: Create pcsd certificate and private_key files
```
Just to be clear about which certificate and key is being created. The role works with several pre-shared keys, and, once #68 merged, TLS certificates for qdevice. So it may get confusing.

tomjelinek (changes requested):
The `tests_cluster_basic_custom_pcsd_tls_cert_role.yml` is failing on my nodes with the following error:
```
The conditional check '__certificate_provider_vars[__certificate_provider].packages is defined                                                                  
' failed. The error was: error while evaluating conditional (__certificate_provider_vars[__certificate_provider].packages is defined                            
): '__certificate_provider' is undefined

The error appears to be in '/root/.ansible/collections/ansible_collections/fedora/linux_system_roles/roles/certificate/tasks/main.yml': line 22, column 7, but may                                                                              
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  block:
    - name: Ensure provider packages are installed
      ^ here
fatal: [rh86-node2]: FAILED! => {"msg": "{% set unique_providers = [] %}\n{% for item in certificate_requests %}\n{%   set _ = unique_providers.append(\n       item.provider |\n       d(__certificate_provider_default)\n     ) %}\n{% endfor %}\n{{ unique_providers | unique }}\n: {% for _cert in ha_cluster_pcsd_certificates %}\n{%   set _ = _cert.__setitem__('name', '/var/lib/pcsd/pcsd') %}\n{% endfor %}\n{{ ha_cluster_pcsd_certificates }}\n: [AnsibleMapping([('name', '{{ __cert_name }}'), ('common_name', '{{ ansible_host }}'), ('ca', 'self-sign')])]: 'ansible_host' is undefined"}
```
I cannot test this PR. I would like to be able to run the tests before I approve the changes. I'm not sure if the problem is in my setup or elsewhere. Since the CI passes, it may be a problem in my setup. But I was unable to figure out what the problem is and how to resolve it. Any help appreciated.

nhosoi:
> The `tests_cluster_basic_custom_pcsd_tls_cert_role.yml` is failing on my nodes with the following error:
> 
> ```
> The conditional check '__certificate_provider_vars[__certificate_provider].packages is defined                                                                  
> ' failed. The error was: error while evaluating conditional (__certificate_provider_vars[__certificate_provider].packages is defined                            
> ): '__certificate_provider' is undefined
> 
> The error appears to be in '/root/.ansible/collections/ansible_collections/fedora/linux_system_roles/roles/certificate/tasks/main.yml': line 22, column 7, but may                                                                              
> be elsewhere in the file depending on the exact syntax problem.
> 
> The offending line appears to be:
> 
>   block:
>     - name: Ensure provider packages are installed
>       ^ here
> fatal: [rh86-node2]: FAILED! => {"msg": "{% set unique_providers = [] %}\n{% for item in certificate_requests %}\n{%   set _ = unique_providers.append(\n       item.provider |\n       d(__certificate_provider_default)\n     ) %}\n{% endfor %}\n{{ unique_providers | unique }}\n: {% for _cert in ha_cluster_pcsd_certificates %}\n{%   set _ = _cert.__setitem__('name', '/var/lib/pcsd/pcsd') %}\n{% endfor %}\n{{ ha_cluster_pcsd_certificates }}\n: [AnsibleMapping([('name', '{{ __cert_name }}'), ('common_name', '{{ ansible_host }}'), ('ca', 'self-sign')])]: 'ansible_host' is undefined"}
> ```
> 
> I cannot test this PR. I would like to be able to run the tests before I approve the changes. I'm not sure if the problem is in my setup or elsewhere. Since the CI passes, it may be a problem in my setup. But I was unable to figure out what the problem is and how to resolve it. Any help appreciated.

Hi @tomjelinek, I'm trying to reproduce the problem. Could you please tell me the os of the control host and the managed host? And the version of ansible? When testing locally, we usually use qemu. How about your case? Thanks!

README.md
nhosoi: In the current behavior, if `ha_cluster_pcsd_public_key_src` and `ha_cluster_pcsd_private_key_src` variables are set along with `ha_cluster_pcsd_certificates`, the first two variables are just ignored. Do we still want to make the case `fail`? It's a bit tricky. It does not do any harm even if all 3 variables are set, but it may surprise the customer if they expect their existing cert and key will not be copied to the managed node. What do you think, @tomjelinek?

README.md
```diff
@@ -194,6 +194,26 @@ https://docs.ansible.com/ansible/latest/user_guide/vault.html for details.
 If these variables are set, `ha_cluster_regenerate_keys` is ignored for this
 certificate - key pair.
 
+#### `ha_cluster_pcsd_certificates`
+
+You can either use existing TLS certificate and private key files by providing
+them with [`ha_cluster_pcsd_public_key_src` and `ha_cluster_pcsd_private_key_src`](#ha_cluster_pcsd_public_key_src-ha_cluster_pcsd_private_key_src),
+or use the role to create certificates by providing `ha_cluster_pcsd_certificates`.
```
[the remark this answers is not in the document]
nhosoi: Thanks, @tomjelinek.

Aside from this pr, if `ha_cluster_pcsd_public_key_src` and `ha_cluster_pcsd_private_key_src` are **not** set, then `pcsd` automatically generates the private key and the cert internally? Did I read your inputs correctly? If so, I'd think there's no much benefit to add "using the `certificate` role" from the `ha_cluster` role, is it?

To reduce the complexity, I can close this pr if you agree. Maybe, would it be preferable just adding the sentence like "If `ha_cluster_pcsd_public_key_src` and `ha_cluster_pcsd_private_key_src` are **not** set, the `ha_cluster` role will create pcsd certificates via pcsd itself"?

tomjelinek:
> Hi @tomjelinek, I'm trying to reproduce the problem. Could you please tell me the os of the control host and the managed host? And the version of ansible?

Sure. Sorry for not providing them right in the first post.

OS of both control and managed nodes is RHEL 8.6.0.

Ansible version:
```
# ansible-playbook --version
[DEPRECATION WARNING]: Ansible will require Python 3.8 or newer on the controller starting with Ansible 2.12. Current version: 3.6.8 (default, Jan 14 2022, 11:04:20) [GCC 8.5.0 20210514 (Red Hat 8.5.0-7)]. This feature will be removed from ansible-core in version 2.12.
 Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
ansible-playbook [core 2.11.12]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible-playbook
  python version = 3.6.8 (default, Jan 14 2022, 11:04:20) [GCC 8.5.0 20210514 (Red Hat 8.5.0-7)]
  jinja version = 2.10.1
  libyaml = True
```

Collection version:
```
# ansible-galaxy collection list
[DEPRECATION WARNING]: Ansible will require Python 3.8 or newer on the controller starting with Ansible 2.12. Current version: 3.6.8 (default, Jan 14 2022, 11:04:20) [GCC 8.5.0 20210514 (Red Hat 8.5.0-7)]. This feature will be removed from ansible-core in version 2.12.
 Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

# /root/.ansible/collections/ansible_collections
Collection                Version
------------------------- -------
ansible.posix             1.4.0
community.general         5.6.0
fedora.linux_system_roles 1.27.0

# /usr/share/ansible/collections/ansible_collections
Collection               Version
------------------------ -------
ansible.posix            1.4.0
community.general        3.8.9
redhat.rhel_system_roles 1.16.2
```

> When testing locally, we usually use qemu. How about your case? Thanks!

I'm developing and testing the role on qemu VMs which I also use for pcs development. I don't create new VMs for running the role tests.

README.md
tomjelinek: > It does not do any harm even if all 3 variables are set, but it may surprise the customer if they expect their existing cert and key will not be copied to the managed node. What do you think, @tomjelinek?

Yes, that's my thinking. The role would not fail, but the outcome may not match the expectations.

The role already has many input variables. It seems quite possible to set `ha_cluster_pcsd_public_key_src` and `ha_cluster_pcsd_private_key_src` and forget to delete `ha_cluster_pcsd_certificates`. In that case, documentation doesn't help. Actively checking and failing the role does.

I'm not sure if such a check is a best practice, though. Until now, `check-and-prepare-role-variables.yml` looks for such variables configurations which would fail the role anyway.

nhosoi:
Thank you, @tomjelinek! At the first glance, I noticed you are using ansible 2.11.
> ansible-playbook [core 2.11.12]

Could it be possible to upgrade it to the newer version? (I'd think at least 2.12 is available.)

README.md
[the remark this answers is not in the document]
nhosoi: Ok. Thank you for the confirmation, @tomjelinek. If there is a benefit, I keep improving this pr. Thanks!

nhosoi:
> Thank you, @tomjelinek! At the first glance, I noticed you are using ansible 2.11.
> 
> > ansible-playbook [core 2.11.12]
> 
> Could it be possible to upgrade it to the newer version? (I'd think at least 2.12 is available.)

Sorry, not an issue of the ansible version :p, but of `ansible_host`. I read "`ansible_host` is a connection variable, which is not included in `hostvars` if it is not explicitly declared." The reason it works for the CI is it is declared in the standard-inventory-qcow2...

Here's the debug print with my reproducer.
```
TASK [redhat.rhel_system_roles.certificate : DEBUG] ************************
task path: /usr/share/ansible/collections/ansible_collections/redhat/rhel_system_roles/role
s/certificate/tasks/main.yml:36
Wednesday 12 October 2022  16:17:03 -0400 (0:00:00.053)       0:00:36.861 ***** 
fatal: [myself]: FAILED! => {}
MSG:
{% set unique_providers = [] %}
{% for item in certificate_requests %}
{%   set _ = unique_providers.append(
       item.provider |
       d(__certificate_provider_default)
     ) %}
{% endfor %}
{{ unique_providers | unique }}
: {% for _cert in ha_cluster_pcsd_certificates %}
{%   set _ = _cert.__setitem__('name', '/var/lib/pcsd/pcsd') %}
{% endfor %}
{{ ha_cluster_pcsd_certificates }}
: [{'name': '{{ __cert_name }}', 'common_name': '{{ ansible_host }}', 'ca': 'self-sign'}]: 
'ansible_host' is undefined
```
Good news is `ansible_hostname` does not have the problem. If I replace `ansible_host` with `ansible_hostname`, the `tests_cluster_basic_custom_pcsd_tls_cert_role.yml` test passes. Let me make the change. Thanks for finding it out!

nhosoi:
[citest]

README.md
```diff
@@ -998,6 +1004,22 @@ to `true` in your playbooks using the `ha_cluster` role.
...
+### Creating pcsd certificate and private key files using the `certificate` role
+
+This example creates self-signed pcsd certificate and private key files
+in /var/lib/pcsd with the file name FILENAME.crt and FILENAME.key, respectively.
+
+```yaml
+- hosts: node1 node2
+  vars:
+    ha_cluster_pcsd_certificates:
+      - name: FILENAME
```
tomjelinek, line 1016: Pcsd only looks for `pcsd.crt` and `pcsd.key` files. The `name` variable is even set by the `ha_cluster` role:
```
{%   set _ = _cert.__setitem__('name', '/var/lib/pcsd/pcsd') %}
```
Therefore, I see no point in letting customers set the `name` variable. I think that the readme should state, that the variable is set automatically and should not be changed.

tomjelinek:
> Good news is ansible_hostname does not have the problem. If I replace ansible_host with ansible_hostname, the tests_cluster_basic_custom_pcsd_tls_cert_role.yml test passes. Let me make the change. Thanks for finding it out!

Thanks for fixing the issue. Now the role passes the previously failing task. However, it still doesn't work for me. It gets stuck at `fedora.linux_system_roles.certificate : Ensure certificate requests` task. I found out that it runs the following command on managed hosts and the command gets stuck:
```
/usr/bin/getcert resubmit -N CN=rh86-node2 -c local -w -f /var/lib/pcsd/pcsd.crt -D  -A  -E  -g 2048 -K  -u keyEncipherment -u digitalSignature -U 1.3.6.1.5.5.7.3.1 -U 1.3.6.1.5.5.7.3.2 -U  -B  -C
```
When I run the command manually, it gives me this error and exits:
```
Error parsing Kerberos principal name "-u": Configuration file does not specify default realm.
```
Certmonger version is `certmonger-0.79.13-5.el8.x86_64`. I don't use kerberos on my VMs. When pcsd certificate and key already exists, the Ansible task says `ok` and tests pass.

nhosoi:
> > Good news is ansible_hostname does not have the problem. If I replace ansible_host with ansible_hostname, the tests_cluster_basic_custom_pcsd_tls_cert_role.yml test passes. Let me make the change. Thanks for finding it out!
> 
> Thanks for fixing the issue. Now the role passes the previously failing task. However, it still doesn't work for me. It gets stuck at `fedora.linux_system_roles.certificate : Ensure certificate requests` task. I found out that it runs the following command on managed hosts and the command gets stuck:
> 
> ```
> /usr/bin/getcert resubmit -N CN=rh86-node2 -c local -w -f /var/lib/pcsd/pcsd.crt -D  -A  -E  -g 2048 -K  -u keyEncipherment -u digitalSignature -U 1.3.6.1.5.5.7.3.1 -U 1.3.6.1.5.5.7.3.2 -U  -B  -C
> ```
> 
> When I run the command manually, it gives me this error and exits:
> 
> ```
> Error parsing Kerberos principal name "-u": Configuration file does not specify default realm.
> ```
> 
> Certmonger version is `certmonger-0.79.13-5.el8.x86_64`. I don't use kerberos on my VMs. When pcsd certificate and key already exists, the Ansible task says `ok` and tests pass.

Hmmm, on my test host from `1minutetip`, I don't have kerberos nor certmonger configured.
```
# rpm -qa | egrep krb5
krb5-libs-1.18.2-14.el8.x86_64
# rpm -qa | egrep certmonger
#
```
This is my task output, which was successful.
```
TASK [redhat.rhel_system_roles.certificate : Ensure certificate requests] ******
task path: /usr/share/ansible/collections/ansible_collections/redhat/rhel_system_roles/roles/certificate/tasks/main.yml:109
Thursday 13 October 2022  12:22:15 -0400 (0:00:01.219)       0:00:49.742 ****** 
changed: [myself] => (item={'name': '/var/lib/pcsd/pcsd', 'common_name': 'ci-vm-10-0-137-248', 'ca': 'self-sign'}) => {
    "ansible_loop_var": "item",
    "changed": true,
    "item": {
        "ca": "self-sign",
        "common_name": "ci-vm-10-0-137-248",
        "name": "/var/lib/pcsd/pcsd"
    }
}
MSG:
Certificate requested (new).
META: role_complete for myself
```
I wonder how I can reproduce the problem... Again, I'm on 8.6 and using ansible 2.12.
```
# cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.6 (Ootpa)
# ansible-playbook --version
ansible-playbook [core 2.12.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible-playbook
  python version = 3.8.12 (default, Sep 16 2021, 10:46:05) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
  jinja version = 2.10.3
  libyaml = True
```

nhosoi:
[citest]

nhosoi:
[citest pending]

tomjelinek:
@nhosoi I merged a branch adding support for `qnetd` and `qdevice`. Communication between those cluster components is TLS encrypted and the role sets up TLS certificates. Could you take a look at that to check if this is something we want to do with the `certificate` role? Thanks.

tomjelinek (approved):

nhosoi:
> @nhosoi, Thank you for helping me with the `getcert` issue! With a bit of experimentation and reading documentation, I was able to figure out what the problem was (see bellow). I fixed my VMs and updated my tests scripts. Your patch now passes all tests on my systems.
> 
> When one creates a certificate and key using `certmonger` (which the `certificate` role does), `certmonger`stores information about the certificate and key in its database. If the certificate and key files are then deleted while keeping the corresponding database records intact (which I did to test the role), then the `getcert` command gets stuck. And so does the Ansible task running it. I managed to fix it by running `getcert stop-tracking --certfile=/var/lib/pcsd/pcsd.crt --keyfile=/var/lib/pcsd/pcsd.key`. That removed the database records related to pcsd certificate and key.
> 
> I imagine this may be difficult to debug for customers using the role and not knowing what it does behind the scenes. If the issue cannot be resolved by the role, I think it should be at least documented. What do you think?

Thank you soooo much for figuring out the issue of `getcert`, @tomjelinek!

I could reproduce the `getcert` hung issue by removing `/var/lib/pcsd/pcsd.key` and `/var/lib/pcsd/pcsd.crt` before calling the 2nd `certificate` role.

Let's ask @rjeffman what's the best way to workaround it...

@rjeffman, could you please advise? We are trying to use the `certificate` role from the `ha_cluster` role. @tomjelinek found out if we create the key and cert pair by the `certificate` role, remove them manually, then run the `certificate` role again, the role hangs in `/usr/bin/getcert resubmit` command line invoked from `module_utils/certificate_lsr/providers/certmonger.py`.

We'd like to learn how we could avoid this hanging. Manual removing the key and cert pair should be avoided? But once it happens, we need some way to reset it... @tomjelinek figured out it's fixed by calling `getcert stop-tracking --certfile=/var/lib/pcsd/pcsd.crt --keyfile=/var/lib/pcsd/pcsd.key`, but it's hard for the other roles that use the `certificate` role to do...

If needed, we can move to https://github.com/linux-system-roles/certificate/issues to continue the discussion. Please let us know. Thanks!!

nhosoi:
> @nhosoi I merged a branch adding support for `qnetd` and `qdevice`. Communication between those cluster components is TLS encrypted and the role sets up TLS certificates. Could you take a look at that to check if this is something we want to do with the `certificate` role? Thanks.

Will do. Thanks! BTW, I rebased on top of your merges and squashed.

rjeffman:
@nhosoi, @tomjelinek, I need a few days to look into it. Unfortunately, not before next Thursday (Oct 18).

Let's keep the discussion here, in one place, at least until we understand it is a certificate role issue (and not something else).

nhosoi:
[citest]

tomjelinek:
> @tomjelinek, I got one question regarding the key and cert pairs for `qnetd` and `qdevice`. You prepared `ha_cluster_pcsd_private_key_src` and `ha_cluster_pcsd_public_key_src` to copy to the pair to the pcsd node. But not for `qnetd` and `qdevice`. Are there any reason why you did not? Is it because it's better rely on the pair generated by `qnetd` and `qdevice`? Thanks!

I think `qnetd` only accepts certificates signed by its certificate authority. More details can be found in `corosync-qdevice(8)` man page, section 'MODEL NET TLS CONFIGURATION', and `corosync-qdevice-net-certutil` script.

Pcs reimplements `corosync-qdevice-net-certutil`, and `ha_cluster` role reimplements that.

Since the certificates are stored in NSS database, a simple copy may not work. It's possible to get a signed certificate from `qnetd` CA, send it to all nodes, and let them import it into their NSS databases (pcs does that). Doing it the same way in the role would just add more code, so I decided it's better to leverage Ansible and run the script for setting certificates up on all nodes.

nhosoi:
> I think `qnetd` only accepts certificates signed by its certificate authority. More details can be found in `corosync-qdevice(8)` man page, section 'MODEL NET TLS CONFIGURATION', and `corosync-qdevice-net-certutil` script.
> 
> Pcs reimplements `corosync-qdevice-net-certutil`, and `ha_cluster` role reimplements that.
> 
> Since the certificates are stored in NSS database, a simple copy may not work. It's possible to get a signed certificate from `qnetd` CA, send it to all nodes, and let them import it into their NSS databases (pcs does that). Doing it the same way in the role would just add more code, so I decided it's better to leverage Ansible and run the script for setting certificates up on all nodes.

Thanks for your inputs, @tomjelinek . Certs are stored in `NSS`... Indeed, it'll be harder or even impossible to make it work with the `certificate` role. Currently, `certomonger` is the only a provider the `certificate` role supports. So, most likely the CA in the NSS database is out of the scope even if it's retrieved and placed as a crt format...

I agree with your decision. Let's finish this task with `pcsd` support only. Could you please merge this pr, @tomjelinek? I'd like to have an advice from @rjeffman regarding the cert removal issue, but it can be done in a separate vehicle (it might be in the `certificate` role). Thanks!

tomjelinek:
> Let's finish this task with `pcsd` support only. I'd like to have an advice from @rjeffman regarding the cert removal issue, but it can be done in a separate vehicle (it might be in the `certificate` role). Thanks!

Sounds good to me.

PR #78, opened 2022-11-17
allow enabled SBD on disabled cluster

yeoldegrove:
Currently the `sbd.service`will not be enabled if the cluster autostart is disabled. This is not intended behavior as is will effectively break the feature.
We can simply remove the condition to depend on `ha_cluster_start_on_boot` as on a RHEL8 system the `sbd.service` has a dependencies (`Before/After/PartOf/RequiredBy`) to cluster related services which make sure it is only ever started by the cluster (a manual start is not possible).

```
# /usr/lib/systemd/system/sbd.service
[Unit]
[...]
Before=pacemaker.service
Before=dlm.service
After=systemd-modules-load.service iscsi.service
PartOf=corosync.service
[...]

[Install]
RequiredBy=corosync.service
RequiredBy=pacemaker.service
RequiredBy=dlm.service
```

### DIFF
--- a/tasks/cluster-enable-disable.yml
+++ b/tasks/cluster-enable-disable.yml
@@ -19,7 +19,7 @@
 - name: Enable or disable SBD
   service:
     name: sbd
-    enabled: "{{ ha_cluster_sbd_enabled and ha_cluster_start_on_boot }}"
+    enabled: "{{ ha_cluster_sbd_enabled }}"
     # Null (i.e. nochange) is currently not supported: Either all nodes are
     # configured to start the cluster on boot or not to start the cluster on
     # boot. Null (nochange) could lead to situations where part of cluster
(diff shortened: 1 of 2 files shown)


yeoldegrove:
In theory `tests/tests_cluster_basic_disabled.yml` would need an adaption. But I did not find any hints that these tests are run or how to run them.

yeoldegrove:
@richm Thanks for the hint to the docs! The test is now adapted and the PR is good to go from my side.

richm:
[citest]

richm:
@yeoldegrove thank you for the PR.  Unfortunately our CI system does not like git branch names with the `/` character in them.  Would it be possible for you to use a different branch name without a `/` in it?  maybe `feat-allow_sbd_and_disabled_cluster`?  and same for https://github.com/linux-system-roles/ha_cluster/pull/79 ?  We're working on a fix, but it will take some time.

yeoldegrove:
@richm #81 is the follow up

PR #87, opened 2022-12-06
tests: add qnetd cleanup

tomjelinek:
Ensure that qnetd package is not installed and there are no qnetd config files present to verify that the role really installs and creates those. This also sets test environment to a clearly defined and correct state, which is needed in case tests are not isolated.

### DIFF
--- a/tasks/test_cleanup_qnetd.yml
+++ b/tasks/test_cleanup_qnetd.yml
@@ -1,0 +1,18 @@
+# SPDX-License-Identifier: MIT
+# This task configures test systems for testing in the CI environment.
+# Only use this task for testing.
+# This task file resides in `tasks/` so that it can be used from other roles
+# with `include_role` with `tasks_from: filename` directly without the need to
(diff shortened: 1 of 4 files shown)


richm:
ok - will it cause a problem for tests that come after the qnetd related tests, if qnetd is running?

tomjelinek:
> ok - will it cause a problem for tests that come after the qnetd related tests, if qnetd is running?

No, I'm not aware of any problems in this scenario. The running qnetd should be silently ignored. And the role de-configures qnetd anyway, as non-qnetd tests don't set `ha_cluster_qnetd.present` to `true`.

richm:
I had to make some fixes for ansible-core 2.14 (get rid of command `warn` argument) please rebase on top of the latest master branch

tomjelinek:
> I had to make some fixes for ansible-core 2.14 (get rid of command `warn` argument) please rebase on top of the latest master branch

Done.

PR #92, opened 2022-12-14
Not request password to be specified when purging cluster

spetrosi:
When running the role with ha_cluster_cluster_present: false to purge cluster passwords are not required

### DIFF
--- a/tasks/check-and-prepare-role-variables.yml
+++ b/tasks/check-and-prepare-role-variables.yml
@@ -7,7 +7,7 @@
         msg: "{{ item }} must be specified"
       when:
         - lookup("vars", item, default="") | string | length < 1
-        - ha_cluster_cluster_present
+        - ha_cluster_cluster_present | bool
       loop:
         - ha_cluster_hacluster_password
       run_once: true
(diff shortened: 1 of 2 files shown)


spetrosi:
[citest]

richm (approved):
lgtm - @tomjelinek ?

tomjelinek (changes requested):
There is the "Set hacluster password" task in `tasks.main.yml`. The task should not be run if `ha_cluster_hacluster_password` is not set. Moving the task to "Install and configure HA cluster" block should do the trick.

spetrosi:
> There is the "Set hacluster password" task in `tasks.main.yml`. The task should not be run if `ha_cluster_hacluster_password` is not set. Moving the task to "Install and configure HA cluster" block should do the trick.

I tried and didn't succeed implementing what you've described. Can you add a commit with what you suggest?

I use the following playbook for testing:

```yaml
- name: Purge cluster and mssql configuration
  hosts: all
  vars:
    ha_cluster_cluster_present: no
    ha_cluster_enable_repos: false
  roles:
```

tomjelinek:
> > There is the "Set hacluster password" task in `tasks/main.yml`. The task should not be run if `ha_cluster_hacluster_password` is not set. Moving the task to "Install and configure HA cluster" block should do the trick.
> 
> I tried and didn't succeed implementing what you've described. Can you add a commit with what you suggest?

Basically, we want to prevent setting `hacluster` password if it is empty / not set. I added a commit for that.

tomjelinek (approved):

richm (approved):

PR #94, opened 2023-01-01
ansible-lint 6.x fixes

richm:
All tasks should be named - added names for some tasks
The `block` should come last in a task block - the `vars` and `when` and
other parameters should come before the `block` and the list of tasks
for readability
Improve jinja spacing for readability
If you use a template in a task name, it must come last


richm:
[citest]

tomjelinek (approved):

PR #95, opened 2023-01-04
more ansible-lint 6.x fixes

richm:
* All tasks and plays must be named - even blocks
* In blocks, all parameters should come before block, then rescue, then always
* Fix some jinja spacing, suppress some warnings
* ansible-lint does not like the `run_once: true` in the tests, so suppress

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,3 +1,20 @@
 ---
 skip_list:
   - fqcn-builtins
+exclude_paths:
+  - tests/roles/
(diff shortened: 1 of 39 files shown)


richm:
[citest]

tomjelinek (approved):
Thanks for the fix! I just changed a few block names.

richm:
[citest]

richm:
[citest]

PR #97, opened 2023-01-10
Add check for non-inclusive language

nhosoi:
Add check for non-inclusive language
    
Add a check for usage of terms and language that is considered
non-inclusive. We are using the woke tool for this with a wordlist
that can be found at
https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/woke.yml

Clean up / Workaround non-inclusive words

- CHANGELOG.md
- README.md
- examples/constraints.yml
- examples/resources.yml
- tasks/create-and-push-cib.yml
- tests/tests_cib_constraints_create.yml
- tests/tests_cib_resources_create.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -1247,13 +1247,13 @@ in /var/lib/pcsd with the file name FILENAME.crt and FILENAME.key, respectively.
                 value: '5'
               - name: interval
                 value: '1min'
-      - id: dmy-1
+      - id: example-1
         # wokeignore:rule=dummy
         agent: 'ocf:pacemaker:Dummy'
-      - id: dmy-2
+      - id: example-2
         # wokeignore:rule=dummy
         agent: 'ocf:pacemaker:Dummy'
-      - id: dmy-3
+      - id: example-3
         # wokeignore:rule=dummy
         agent: 'ocf:pacemaker:Dummy'
       - id: simple-clone
@@ -1268,8 +1268,8 @@ in /var/lib/pcsd with the file name FILENAME.crt and FILENAME.key, respectively.
     ha_cluster_resource_groups:
       - id: simple-group
         resource_ids:
-          - dmy-1
-          - dmy-2
+          - example-1
+          - example-2
         meta_attrs:
           - attrs:
               - name: target-role
@@ -1278,7 +1278,7 @@ in /var/lib/pcsd with the file name FILENAME.crt and FILENAME.key, respectively.
                 value: 'true'
       - id: cloned-group
         resource_ids:
-          - dmy-3
+          - example-3
     ha_cluster_resource_clones:
       - resource_id: simple-clone
       - resource_id: clone-with-options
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -1243,23 +1247,29 @@ in /var/lib/pcsd with the file name FILENAME.crt and FILENAME.key, respectively.
                 value: '5'
               - name: interval
                 value: '1min'
-      - id: dummy-1
+      - id: dmy-1
```
richm, line 1250: `dmy` is still close to `dummy` (although it does pass `woke`).  I would prefer a better name like `resource-1` or `example-1` or something like that.

examples/constraints.yml
```diff
@@ -16,60 +16,82 @@
           - attrs:
               - name: pcmk_host_list
                 value: node1 node2
-      - id: dummy-1
+      # wokeignore:rule=dummy
```
richm, line 19: do you need this wokeignore?  you have one below for the `ocf:pacemaker:Dummy`

richm: There are several other places below where it looks as though you have a wokeignore for the `dmy`?

tests/tests_cib_constraints_create.yml
```diff
@@ -546,24 +559,28 @@
 
         - name: Verify colocation constraints
           vars:
+            # wokeignore:rule=master
+            __mstr: 'Master'
```
richm, line 563: you don't need the quotes here

tests/tests_cib_constraints_create.yml
```diff
@@ -424,22 +429,30 @@
         # yamllint disable rule:line-length
         - name: Verify location constraints
           vars:
+            # wokeignore:rule=master
+            __um: Master
```
richm, line 433: good idea - this will make it easier to manage

tests/tests_cib_constraints_create.yml
```diff
@@ -620,18 +637,22 @@
 
         - name: Verify ticket constraints
           vars:
+            # wokeignore:rule=master
+            __mstr: 'Master'
```
richm, line 641: you don't need the quotes here

tests/tests_cib_resources_create.yml
```diff
@@ -61,14 +61,17 @@
               - name: interval
                 value: '28s'
 
-      - id: dummy-no-attrs
+      - id: dmy-no-attrs
```
richm, line 64: would prefer not to use `dmy` here - maybe `resource-no-attrs` or `test-no-attrs` or similar

tests/tests_cib_resources_create.yml
```diff
@@ -514,23 +530,29 @@
           when:
             - '"pcmk.resource.config.output-formats" not in __test_pcs_capabilities'
           vars:
+            # wokeignore:rule=master
+            __mstr: 'Master'
```
richm, line 534: you don't need quotes here

tests/tests_cib_resources_create.yml
```diff
@@ -767,26 +789,32 @@
                and
                ansible_facts['distribution_major_version'] == '8') |
                bool }}"
-            __promoted: "{{ __old_roles | ternary('Master', 'Promoted') }}"
-            __unpromoted: "{{ __old_roles | ternary('Slave', 'Unpromoted') }}"
+            # wokeignore:rule=master
+            __mstr: 'Master'
```
richm, line 793: you don't need quotes here

tests/tests_cib_resources_create.yml
```diff
@@ -514,23 +530,29 @@
           when:
             - '"pcmk.resource.config.output-formats" not in __test_pcs_capabilities'
           vars:
+            # wokeignore:rule=master
+            __mstr: 'Master'
+            # wokeignore:rule=slave
+            __slv: 'Slave'
+            # wokeignore:rule=dummy
+            __dmy: 'Dummy'
```
richm, line 538: I think using `__dmy` here is ok (as opposed to other places where you replaced the string "dummy" with "dmy" - those places should use "test" or "example" or "resource" or similar instead of "dmy") - in this case it is good because it is a replacement for places where you must use the explicit string `Dummy` and it signifies that

tomjelinek (approved):
Thank you for the patch!

richm:
[citest]

PR #105, opened 2023-02-02
Fix stonith watchdog timeout

tomjelinek:
Fix a corner case bug occuring with pcs 0.10.15 and newer and pcs 0.11.4 and newer:
When setting `stonith-watchdog-timeout` in a stopped cluster, `cib.xml.sig` is not modified by pcs. This leads to pacemaker ignoring new `cib.xml` content and read configuration from a previous version of cib, effectively reverting `stonith-watchdog-timeout` update done by the role. Pacemaker then exits with an error and the role fails, unable to proceed and configure the cluster.

With old pcs versions, removing `cib.xml.sig` has no adverse effects. There is no need to check for pcs version when removing the file.

### DIFF
--- a/tasks/sbd.yml
+++ b/tasks/sbd.yml
@@ -63,28 +63,39 @@
   service_facts:
 
 - name: Set stonith-watchdog-timeout cluster property
-  command:
-    cmd: >
-      pcs --force
-      {{
-        (
-          ansible_facts.services['pacemaker.service']['state'] | d()
-          == 'running'
-        ) | ternary('', '-f /var/lib/pacemaker/cib/cib.xml')
-      }}
-      -- property set
-      stonith-watchdog-timeout={{ ha_cluster_sbd_enabled | ternary('', '0') }}
-  changed_when: true
+  vars:
+    pacemaker_running: "{{
+          ansible_facts.services['pacemaker.service']['state'] | d('')
+          == 'running' }}"
+  block:
+    - name: Set stonith-watchdog-timeout cluster property in CIB
+      command:
+        cmd: >
+          pcs --force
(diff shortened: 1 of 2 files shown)


.github/workflows/ansible-managed-var-comment.yml
```diff
@@ -1,4 +1,5 @@
 ---
+# yamllint disable rule:line-length
```
richm, line 2: This shouldn't be necessary anymore

tasks/sbd.yml
```diff
@@ -63,28 +63,41 @@
...
-  command:
-    cmd: >
-      pcs --force
-      {{
-        (
-          ansible_facts.services['pacemaker.service']['state'] | d()
+  vars:
+    pacemaker_running: "{{
+          ansible_facts.services['pacemaker.service']['state'] | d('')
           == 'running'
```
richm, line 69:
```suggestion
          == 'running' }}"
```

tasks/sbd.yml
```diff
@@ -63,28 +63,41 @@
...
-        ) | ternary('', '-f /var/lib/pacemaker/cib/cib.xml')
-      }}
-      -- property set
-      stonith-watchdog-timeout={{ ha_cluster_sbd_enabled | ternary('', '0') }}
-  changed_when: true
+      }}"
```
richm, line 70:
```suggestion
```
Ansible prefers to have the ` }}"` at the end of the line with code on it, rather than on a line by itself

richm:
[citest]

richm:
[citest bad]

richm (approved):

PR #106, opened 2023-02-02
Fence agent firewall port is restricted to x86_64 architecture.

nhosoi:
Add the condition to tests/tasks/check_firewall_selinux.yml and tasks/selinux.yml.


nhosoi:
[citest]

richm:
I verified this fix works on aarch64

nhosoi:
> I verified this fix works on aarch64

Thank you soooooo much, @richm !!

PR #116, opened 2023-03-29
Fingerprint RHEL System Role managed config files

nhosoi:
Add role name to the generated config files. 
```
# system_role:ha_cluster
```
Note: This info is introduced to help customers to identify that it was generated by System Roles. Also, it could be used in future analysis.

### DIFF
--- a/templates/sbd
+++ b/templates/sbd
@@ -1,4 +1,5 @@
 {{ ansible_managed | comment }}
+{{ "system_role:ha_cluster" | comment(prefix="", postfix="") }}
 
 {% macro option(ansible_name, default_value) -%}
   {%- for option in options if option.name == ansible_name -%}


tomjelinek:
[citest]

tomjelinek (approved):

richm:
The fingerprint should be added to the other files managed by the role https://github.com/linux-system-roles/ha_cluster/tree/main/files - note that these files are not generated from a template, so they don't have the `ansible_managed` header/comment, but they are still managed by the role, and should have the fingerprint.

There are other roles which have `files/`, such as nbde_client, firewall, and mssql.

richm (approved):
The `files/` are only used by the `script` module and do not need a header.

PR #118, opened 2023-04-07
fix ansible-lint issues with changed_when, others

richm:
Ansible lint now requires the use of `changed_when:` with commands
even in some cases where the command has a `when` condition.


richm:
[citest]

Issue #122, opened 2023-04-25
Extend platform/version execution to allow either pcs or crmsh


sean-freeman:
# Extend platform/version execution to allow either pcs or crmsh

By design, `ha_cluster` allows for additional platform/version. The first Ansible Task executed within the Ansible Role, is [`set_vars.yml`](https://github.com/linux-system-roles/ha_cluster/blob/main/tasks/set_vars.yml) to import different variables for OS major.minor versions.

An extension of this principle to also include/execute a different set of Ansible Tasks for different high-level command interfaces to Linux Pacemaker, would be valuable to future compatibility.

## Suggestion for execution logic fork

1. Append new var `_ha_cluster_pacemaker_shell` to all existing `/vars/<<os_version>>.yml` files. For example:
 
**`/vars/RHEL9.yml`**
```
_ha_cluster_pacemaker_shell: pcs
```

<br/>

2. Create new directory for Ansible Tasks of a specific shell for Linux Pacemaker. Move the existing specific files of pcs shell into `/tasks/pcs` directory, and create stub code for crmsh shell in future. For example:
 
**`Directory re-structure....`**
```
/tasks/pcs/cluster-start-and-reload.yml
etc.

/tasks/crmsh/cluster-start-and-reload.yml
etc.
```

<br/>

3. Move common Ansible Tasks that execute the underlying Pacemaker binaries `cibadmin` and `crm_mon` into a separate tasks subdirectory. For example:
 
**`Directory re-structure....`**
```
/tasks/common/create-and-push-cib.yml
/tasks/common/cluster-start-and-reload.yml
```

<br/>

4. Append include_tasks. For example

**`/tasks/main.yml`**
```
---
- name: Set platform/version specific variables
  include_tasks: set_vars.yml

- name: Check and prepare role variables
  include_tasks: {{ pacemaker_shell }}/check-and-prepare-role-variables.yml
...
...
```

## Side benefits

- Retains flat design of `/tasks` directory, but reduces some of the contents so it is easier to see the key/controlling Ansible Tasks files and what they execute

---

## Comparison of `pcs` and `crmsh`

Looking in the repository, there are a limited number of `pcs` shell commands in use. See equivilant below for `crmsh`:

```
# PCS Shell                   # CRMSH

pcs cluster                   crm node
pcs constraint <type>         crm configure <type>
pcs property                  crm configure property
pcs qdevice                   crm cluster init qdevice
[...]

PR #123, opened 2023-04-27
refactor: append pacemaker shell option for #122

sean-freeman:
Append pacemaker shell option for #122

This purposefully is light on changes to avoid regression concerns:
- add variable `ha_cluster_pacemaker_shell` to defaults, and use this var in the include_tasks paths
- use existing `set_vars.yml` to set variables for OS Family
- move all Ansible Task YAML files using binary `pcs` into subdirectory `/tasks/shell_pcs`

The PR adds placeholders for `crmsh`, but this PR focuses only on appending the option and execution logic fork.... it does not include the `crmsh` execution logic.


richm:
Ignore the `woke` failure - it is a known bug.

commitlint - your commit message and PR message must be formatted in Conventional Commits format described at https://linux-system-roles.github.io/contribute.html "Write a good commit message" - I suggest just adding `refactor: ` before your commit title line and PR title line e.g.
```
refactor: Append pacemaker shell option for #122

Append pacemaker shell option for https://github.com/linux-system-roles/ha_cluster/issues/122
...
```

I'll comment about the ansible-lint changes in-line

tasks/main.yml
```diff
@@ -18,7 +18,7 @@
     state: present
 
 - name: Check and prepare role variables
-  include_tasks: check-and-prepare-role-variables.yml
+  include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml
```
richm, line 21:
```suggestion
  include_tasks: shell_{{
    ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml
```
or
```
include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/\
  check-and-prepare-role-variables.yml
```
or 
```
include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml  # noqa yamllint[line-length]
```
or see https://linux-system-roles.github.io/documentation/howto/working-with-ansible-jinja-code-and-filters.html

richm:
I realize the line length restrictions may seem onerous - but we're really trying to abide by Ansible recommendations, and use the standard Ansible tools such as `ansible-lint` with their default settings.  These tools are also used at various gating points such as import into Galaxy and Automation Hub.

richm:
sent you a PR to address the remaining ansible-lint issues

richm:
[citest]

richm:
[citest bad]
I cannot reproduce the failures locally - transient problem?

richm:
> [citest bad] I cannot reproduce the failures locally - transient problem?

Yes - there is a problem with the CI system with 9.3 HA repos - please ignore the failures

This PR lgtm - @tomjelinek ?

defaults/main.yml
```diff
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: MIT
 ---
+
+# Linux Pacemaker shell, pcs or crmsh
+ha_cluster_pacemaker_shell: pcs
```
tomjelinek, line 5: I would drop 'pacemaker' here, as pcs handles also corosync, sbd and other cluster components. Crmsh should be able to configure other components than pacemaker, too, I think.

tomjelinek (approved):
Newly created files are missing SPDX-License-Identifier. I'm not sure how big of a problem it is, though.

Apart from the tree mentioned minor issues this looks good to me. Thank you for the PR.

defaults/main.yml
sean-freeman: @tomjelinek I was following nomenclature from elsewhere in Linux Pacemaker projects. I don't mind altering this var name, but from an end user perspective they would refer to 'use Linux Pacemaker' because that is the owning name of all subprojects, sideprojects and components.

For example....

`"pcs and crm shell are two popular higher-level command-line interfaces to Pacemaker."`
https://clusterlabs.org/pacemaker/doc/2.1/Pacemaker_Administration/html/pcs-crmsh.html

`"crmsh is a cluster management shell for the Pacemaker High Availability stack."`
https://crmsh.github.io/

sean-freeman: Also, I can add the identifier to all new (mostly placeholder) files.

tomjelinek: @sean-freeman That makes sense. Let's keep it the `ha_cluster_pacemaker_shell` then.

richm:
Any further objections?  Is this ready to merge?

tomjelinek:
> Any further objections? Is this ready to merge?

I don't have any further objections.

PR #129, opened 2023-05-15
refactor: Update for new pcs

tomjelinek:
refactor: update qdevice related code for upcoming pcs release
test: update tests for upcoming pcs release
style: fix linter issues in SBD task files


tomjelinek:
Regarding CI complaining about Python issues:

- The python modules in the role depend on pcs. Pcs requires Python 3.6. Therefore, there is no point in making the modules compatible with Python 2.7 and 3.5.
- For the same reason, there is no point in adding `from __future__ import` and `__metaclass__ = type`. But since that would be a noop and easy to add to source, I can do that if that's easier than convincing ansible-lint to ignore those.
- `ModuleNotFoundError: No module named 'dacite'` - dacite is distributed with pcs and therefore it is available on managed nodes

I'll take a look at the other reported errors.

richm:
@tomjelinek re: ansible-test failures - see for example https://github.com/linux-system-roles/storage/ and the various `.sanity-ansible-ignore-2.x.txt` files - this is where you can suppress errors with `import`, `compile`, etc.

> The python modules in the role depend on pcs. Pcs requires Python 3.6. Therefore, there is no point in making the modules compatible with Python 2.7 and 3.5.

+1

> For the same reason, there is no point in adding from __future__ import and __metaclass__ = type. But since that would be a noop and easy to add to source, I can do that if that's easier than convincing ansible-test to ignore those.

It's easier to just add these to the python code.  That's what we did for other system roles.

richm:
I'll need to add ha_cluster to the list of roles that provide python code at https://github.com/linux-system-roles/.github/blob/main/inventory/inventory.yml#L132 - that will add the github actions such as codeql and python unit tests.

richm:
@tomjelinek added python checks - you'll need to rebase to pick them up https://github.com/linux-system-roles/ha_cluster/pull/131

tasks/shell_pcs/cluster-setup-keys.yml
```diff
@@ -34,18 +34,29 @@
...
   changed_when: true
 
-- name: Obtain and distribute qdevice certificates
+- name: Remove qdevice certificates [API]
+  pcs_api_v2:
+    cmd_name: qdevice.client_net_destroy
+  when:
+    - not ansible_check_mode
+    - __ha_cluster_qdevice_model != "net" or ha_cluster_regenerate_keys
+    - ('corosync.quorum.device.client.model.net.certificates.rest-api.v2'
```
spetrosi, line 53: Why do you use brackets here? I think using double quotes instead of brackets is conventional in this case.

tasks/shell_pcs/cluster-setup-keys.yml
```diff
@@ -55,9 +66,18 @@
     - __ha_cluster_qdevice_certs.rc == 0
     - __ha_cluster_qdevice_certs.stdout_lines[-1]
       != "** certificate already present **"
-  vars:
-    use_pcs: "{{ 'corosync.quorum.device.client.model.net.certificates.local'
-        in __ha_cluster_pcs_capabilities }}"
   when:
     - __ha_cluster_qdevice_in_use
     - __ha_cluster_qdevice_model == "net"
+    - ('corosync.quorum.device.client.model.net.certificates.local'
```
spetrosi, line 72: brackets here too

tasks/shell_pcs/cluster-setup-keys.yml
```diff
@@ -55,9 +66,18 @@
...
+      not in __ha_cluster_pcsd_capabilities)
+
+- name: Obtain and distribute qdevice certificates [API]
+  pcs_qdevice_certs:
+    qnetd_host: "{{ __ha_cluster_qdevice_host }}"
+    cluster_name: "{{ ha_cluster_cluster_name }}"
+  when:
+    - __ha_cluster_qdevice_in_use
+    - __ha_cluster_qdevice_model == "net"
+    - ('corosync.quorum.device.client.model.net.certificates.local'
```
spetrosi, line 82: brackets here too

tomjelinek:
@richm  No matter what I do, I'm unable to figure out how to install pcs so that it can be imported in Python Unit Tests GitHub actions. Even though the Python Unit Tests action logs show, that both pcs and python3-dacite is being installed, I'm still getting `ModuleNotFoundError: No module named 'pcs'` error. For the unit tests to be able to import dacite, I had to put dacite to requirements.txt. This should not be needed as pcs depends on and installs python3-dacite. Any hints? Note that pcs cannot be installed using pip.

tomjelinek:
@richm Currently, Ubuntu 20.04 and 22.04 are the only available test runners for GitHub actions. However, none of them provide new enough pcs. The python modules in the role require pcs-0.11.4 or newer, whereas those Ubuntu releases ship pcs-0.10.4 and pcs-0.10.11, respectively. So I went ahead and modified the workflow to compile pcs from upstream. As a nice side effect, this allows us to test against various pcs versions. Pcs cannot be `make install`ed, though, as writing to system paths seems to be disallowed in GitHub actions. That shouldn't matter for the tests, though. As long as `PYTHONPATH` is set to point to pcs directory, the tests and modules should be able to import pcs just fine. However, I'm still getting `ModuleNotFoundError: No module named 'pcs'` error from tox and I'm stuck there.

richm:
> @richm Currently, Ubuntu 20.04 and 22.04 are the only available test runners for GitHub actions. However, none of them provide new enough pcs. The python modules in the role require pcs-0.11.4 or newer, whereas those Ubuntu releases ship pcs-0.10.4 and pcs-0.10.11, respectively. So I went ahead and modified the workflow to compile pcs from upstream. As a nice side effect, this allows us to test against various pcs versions. Pcs cannot be `make install`ed, though, as writing to system paths seems to be disallowed in GitHub actions. That shouldn't matter for the tests, though. As long as `PYTHONPATH` is set to point to pcs directory, the tests and modules should be able to import pcs just fine. However, I'm still getting `ModuleNotFoundError: No module named 'pcs'` error from tox and I'm stuck there.

Do we need to test the `pcs` module in the unit tests?  My first thought would be to mock the `pcs` module.  If it is too difficult to mock all of the calls made in that module, then you should be able to use the pcs installed in the system with `make install` by adding the `--sitepackages` argument to tox.  Then, it should work, but only with the system python i.e. the python that `make install` used - that is, if `make install` used system python 3.9, then `tox --sitepackages -e py39` should work, but `-e py38`, `-e py310`, etc. will not work.

tomjelinek:
> you should be able to use the pcs installed in the system with `make install` by adding the `--sitepackages` argument to tox. Then, it should work, but only with the system python i.e. the python that `make install` used - that is, if `make install` used system python 3.9, then `tox --sitepackages -e py39` should work, but `-e py38`, `-e py310`, etc. will not work.

I already tried that. I even modified the workflow to use system python only and defined a matrix to make the workflow run on several versions of Ubuntu, so that multiple python versions are used for the test. The problem is that `make install` fails with an error:
```
+ make install
make  install-recursive
make[1]: Entering directory '/home/runner/work/ha_cluster/ha_cluster/pcs-upstream'
Making install in pcs
make[2]: Entering directory '/home/runner/work/ha_cluster/ha_cluster/pcs-upstream/pcs'
make[3]: Entering directory '/home/runner/work/ha_cluster/ha_cluster/pcs-upstream/pcs'
make[3]: Nothing to be done for 'install-exec-am'.
 /usr/bin/mkdir -p '/usr/share/bash-completion/completions/'
 /usr/bin/install -c -m 644 bash_completion/pcs '/usr/share/bash-completion/completions/'
 /usr/bin/mkdir -p '/usr/share/snmp/mibs'
 /usr/bin/install -c -m 644 snmp/mibs/PCMK-PCS-MIB.txt snmp/mibs/PCMK-PCS-V1-MIB.txt '/usr/share/snmp/mibs'
 /usr/bin/mkdir -p '/etc/default'
 /usr/bin/install -c -m 644 snmp/conf/pcs_snmp_agent '/etc/default'
/usr/bin/install: cannot create regular file '/etc/default/pcs_snmp_agent': Permission denied
make[3]: Leaving directory '/home/runner/work/ha_cluster/ha_cluster/pcs-upstream/pcs'
make[3]: *** [Makefile:859: install-dist_sysconfigDATA] Error 1
make[2]: Leaving directory '/home/runner/work/ha_cluster/ha_cluster/pcs-upstream/pcs'
make[2]: *** [Makefile:951: install-am] Error 2
make[1]: Leaving directory '/home/runner/work/ha_cluster/ha_cluster/pcs-upstream'
make[1]: *** [Makefile:572: install-recursive] Error 1
make: *** [Makefile:879: install] Error 2
Error: Process completed with exit code 2.
```

tomjelinek:
Ok, so I managed to `make install` pcs in a custom directory and point python to it. Then I enabled `sitepackages` in `tox.ini` (I'm not sure if I did it the right way, though) and tests passed.

I'm going to address comments by spetrosi.

tasks/shell_pcs/cluster-setup-keys.yml
tomjelinek: It was my first idea to use brackets. I'll remember using double quotes next time. Thanks.

tomjelinek:
[citest]

richm:
[citest bad]

richm:
the rhel 8.9 ansible-2.9 failure in tests_cib_properties_one_set.yml is reproducible locally:
```
TASK [Check cluster properties configuration] **********************************
task path: /home/rmeggins/linux-system-roles/ha_cluster/tests/tests_cib_properties_one_set.yml:43
Wednesday 24 May 2023  18:11:30 -0600 (0:00:00.051)       0:03:07.553 ********* 
fatal: [/home/rmeggins/.cache/linux-system-roles/rhel-8-9.qcow2]: FAILED! => {
    "assertion": "'no-quorum-policy: stop' in stdout_lines",
    "changed": false,
    "evaluated_to": false
}

MSG:

Assertion failed
```

tomjelinek:
[citest]

richm:
I think whatever the failures are, they are not related to this PR.

PR #134, opened 2023-06-15
feat: cluster and quorum can have distinct passwords

DonEstefan:
Enhancement: cluster and quorum can have distinct passwords

Reason: The same quorum hosts can be joined to multiple, separate clusters. Since the passwords of the "hacluster" user on these clusters might be different, it makes sense to configure them separately.

Result: I separated `pcs auth` for the cluster and qdevice. I also introduced an new var `ha_cluster_hacluster_qdevice_password`, which is used for qdevice auth. If the new var is not set, qdevice auth will fall back to use `ha_cluster_hacluster_password` like rest of the cluster. This should make the change fully backwards compatible. 

Issue Tracker Tickets (Jira or BZ if any): none

### DIFF
--- a/tasks/shell_pcs/pcs-auth-pcs-0.10.yml
+++ b/tasks/shell_pcs/pcs-auth-pcs-0.10.yml
@@ -24,7 +24,8 @@
       {% if __ha_cluster_qdevice_pcs_address %}
         addr={{ __ha_cluster_qdevice_pcs_address | quote }}
       {% endif %}
-    stdin: "{{ ha_cluster_hacluster_qdevice_password | default(ha_cluster_hacluster_password) }}"
+    stdin: "{{ ha_cluster_hacluster_qdevice_password |
+      d(ha_cluster_hacluster_password) }}"
   when:
     - __ha_cluster_qdevice_model == "net"
     - __ha_cluster_qdevice_host
--- a/tasks/shell_pcs/pcs-auth-pcs-0.10.yml
+++ b/tasks/shell_pcs/pcs-auth-pcs-0.10.yml
@@ -30,7 +30,7 @@
       d(ha_cluster_hacluster_password) }}"
   when:
     - __ha_cluster_qdevice_model == "net"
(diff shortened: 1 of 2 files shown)


tasks/shell_pcs/pcs-auth-pcs-0.10.yml
```diff
@@ -1,5 +1,3 @@
-# SPDX-License-Identifier: MIT
----
```
richm, line 2: please put back the license header and the `---` yaml doc start

richm:
[citest]

tasks/shell_pcs/pcs-auth-pcs-0.10.yml
```diff
@@ -12,15 +12,25 @@
...
+      pcs host auth -u hacluster --
+      {{ __ha_cluster_qdevice_host | quote }}
+      {% if __ha_cluster_qdevice_pcs_address %}
+        addr={{ __ha_cluster_qdevice_pcs_address | quote }}
+      {% endif %}
+    stdin: "{{ ha_cluster_hacluster_qdevice_password |
+      d(ha_cluster_hacluster_password) }}"
+  when:
+    - __ha_cluster_qdevice_model == "net"
+    - __ha_cluster_qdevice_host
```
richm, line 33:
```suggestion
    - __ha_cluster_qdevice_host | length > 0
```
this is causing the failures on 2.9 - it doesn't know how to evaluate this in a "bare" context - but 2.15 does.
What is this testing for?  I think this wants to see if `__ha_cluster_qdevice_host` is a non-zero length string since it is set here - https://github.com/linux-system-roles/ha_cluster/blob/main/tasks/shell_pcs/check-and-prepare-role-variables.yml#L58 - it looks like if it wasn't set to a valid host, the value will be `''` the empty string.

tomjelinek:
[citest]

tomjelinek (approved):
Thank you for the patch! It looks good to me. Let's see if it passes CI.

PR #166, opened 2023-12-01
fix: manage firewall on qnetd hosts

### LINKED ISSUE #155: high-availability firewall service is not added on qdevice node
The  ha_cluster_manage_firewall: true attribute does not alter the firewalld configuration for the qdevice node.

In tasks/main.yml in task "Install and configure HA cluster" the firewall.yml inclusion only applies when the ha_cluster_cluster_present is true which will always be false for the qdevice.

Looks like the firewall.yml inclusion should also be added in tasks/shell_pcs/pcs-qnetd.yml

Available firewalld services in qdevice after running the role.

[root@qdevice ~]# 
[root@qdevice ~]# firewall-cmd --list-services
cockpit dhcpv6-client ssh
[root@qdevice ~]#

tomjelinek:
resolves #155

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -35,6 +35,15 @@
 - name: Configure pcs / pcsd
   include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/pcs-configure-pcs-pcsd.yml  # yamllint disable-line rule:line-length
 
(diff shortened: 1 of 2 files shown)


tomjelinek:
[citest]

richm (approved):

PR #173, opened 2023-12-12
test: fix jinja templates in conditionals

tomjelinek:
Fix the error reported by recent Ansible versions: Conditional statements should not include jinja2 templating delimiters. Conditional is marked as unsafe, and cannot be evaluated.


tomjelinek:
[citest]

richm (approved):

PR #177, opened 2024-01-04
ci: Use supported ansible-lint action; run ansible-lint against the collection

richm:
The old ansible-community ansible-lint is deprecated.  There is a
new ansible-lint github action.  The new ansible-lint has several checks
related to ansible-test and the ignore files.  Many of our ignore settings
are not allowed any more and are required to be fixed or addressed in the
Ansible preferred way.

The python imports have to be wrapped in a try/except ImportError, and
where possible, an error must be returned from the module explaining
what was not able to be imported.

The module documentation must comply with the Ansible standards.

Because the module code uses mypy, we have to tell ansible-test
to ignore some code with `compile-2.7!skip` and `import-x.y!skip`,
which ansible-lint does not like, so we have to add
```
skip_list:
  - sanity[cannot-ignore]
```
to make ansible-lint happy.

The latest Ansible repo gating tests run ansible-lint against
the collection format instead of against individual roles.
We have to convert the role to collection format before running
ansible-test.

Role developers can run this locally using
`tox -e collection,ansible-lint-collection`
See https://github.com/linux-system-roles/tox-lsr/pull/125

Add `---` doc start to .markdownlint.yaml


richm:
[citest]

richm:
[citest]

Issue #175, opened 2023-12-20
SUSE Support and crmsh


marcelmamula:
Hello Team,

I am working on [sap-linuxlab](https://github.com/sap-linuxlab) ([community.sap_install](https://github.com/sap-linuxlab/community.sap_install)) and our plan is to make sure that role sap_ha_pacemaker_cluster, which consumes fedora.linux_system_roles, is correctly working on SUSE systems.

I noticed that groundwork for adoption of non-pcs steps was already done thanks to Sean in #122, so adoption would consist of:
- splitting of main.yml based on os_family, because it contains pcs related tasks directly (non-generic names)
- creating variables for Suse
- creating separate tasks under shell_crmsh for crmsh, corosync, sbd
- replicating cluster setup for SUSE from [sap-linuxlab/community.sles-for-sap/cluster](https://github.com/sap-linuxlab/community.sles-for-sap/tree/main/roles/cluster)

There are few things that I wanted to ask for clarification, before proceeding with any changes in fork:
1. Are there any ongoing efforts for SUSE support, or is this free range to pickup?
2. ha_cluster and idempotency?

richm:
>  Are there any ongoing efforts for SUSE support, or is this free range to pickup?

Not that I know of - @sean-freeman or @berndfinger might know.

>    ha_cluster and idempotency?

The role is not currently idempotent for all operations - not sure which are idempotent and which are not - I don't know if you can make the crmsh implementation idempotent without making the pcs implemetation idempotent

sean-freeman:
@richm I had started porting some code in line with the mentioned Ansible Role draft (keeping to the scope of `ha_cluster` to accept config and execute, i.e. not adding any of the commands for a specific platform). However there were other priorities in the SAP LinuxLab Open-Source initiative that needed attention first, so I would probably classify this as free range.

I would probably suggest that `ha_cluster` is psuedo-idempotent in the sense that, if you execute with the same variables you will arrive at the same outcome. However if on multiple runs some variable is changed, it will alter the config file, and the cibadmin will (without mercy or checks) replace any existing config with that new definition.

marcelmamula:
Thank you both for your comments, I will proceed with with work on my end to see how this can be implemented.

tomjelinek:
> if you execute with the same variables you will arrive at the same outcome. However if on multiple runs some variable is changed, it will alter the config file, and the cibadmin will (without mercy or checks) replace any existing config with that new definition.

[...]

PR #186, opened 2024-02-07
crmsh workflow and SUSE support

marcelmamula:
**Enhancement:** This PR adds support for crmsh shell and covers most of its functionalities, with few exceptions.
List of changes:
1. crmsh workflow added, with reusing existing pcs steps where possible and creating rest from scratch, creating jinja2 for each crmsh command.
2. main.yml was adjusted to allow easier loading of shell specific files with removing pcs names and simplifying. 
3. Few pcs shell files were renamed to go along with point 2.
4. Create and push CIB was changed to run_once: true, because running CIB configuration steps on multiple nodes can cause corruption and misaligned CIB files. This is 100% case for crmsh and it would be best if pcs workflow was also cautious of this.
5. corosync.j2 was added for corosync conf creation.
6. crmsh steps are using Maintenance Mode for cluster to freeze config to avoid errors due to versioning caused by cluster resource status changes.

Following functionalities were not part of current scope:
- qnet/qdevice
- resource bundles

**Reason:** Support for SUSE systems and allow wider use, including community.sap_install which is being enabled for SUSE as well.

**Result:** Testing was successfully completed on RHEL9.2 and SLES for SAP 15 SP5 on AWS. It includes:
- Running on fresh servers
- Destroying clusters
- Re-executing and checking results of idempotency
- Consuming using community.sap_install
- local Ansible Lint checks through development

**Issue Tracker Tickets (Jira or BZ if any):** https://github.com/linux-system-roles/ha_cluster/issues/175


tasks/main.yml
```diff
@@ -93,6 +92,9 @@
 
     - name: Create and push CIB
       include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/create-and-push-cib.yml  # yamllint disable-line rule:line-length
+      # CIB changes should be done only on one of cluster nodes to avoid corruption and
+      # inconsistency of resulting cibadmin patch file.
+      run_once: true
```
sean-freeman, line 97: @tomjelinek This looks like a good catch to use `run_once`. Can you confirm this logic amendment? I can see use of `run_once` in many other Ansible Tasks so this change makes sense to me?

tasks/shell_crmsh/create-and-push-cib.yml
```diff
@@ -0,0 +1,266 @@
...
+    suffix: _ha_cluster_cib_xml
+  register: __ha_cluster_tempfile_cib_xml
+  check_mode: false
+  changed_when: not ansible_check_mode
+
+# Maintenance mode is required, because CIB version changes with cluster
+# status changes, resulting in shadow CIB outdated and unable to patch.
+# Sleep is implemented to ensure that cluster have enough time to freeze
+# to ensure CIB export consistency.
+- name: Put cluster in maintenance mode to freeze cib changes
```
sean-freeman, line 34: @tomjelinek This also seems like good logic to follow; even if the Ansible Role was executed on a Linux Pacemaker Cluster running atop hosts for 'PROD' of something, putting into maintenance mode is a industry-wide best practice for any in-place software re-configuration/update?

Is this used elsewhere? As an end-user, I would expect commonality of approach between the different PCMK Shells - although it's not a deal breaker as there will always be some variance between the PCMK Shells.

tasks/main.yml
```diff
@@ -57,8 +57,7 @@
           +
           ha_cluster_sbd_enabled | ternary(__ha_cluster_sbd_packages, [])
           +
-          ha_cluster_fence_agent_packages
-        }}"
+          ha_cluster_fence_agent_packages}}"
```
richm, line 60:
```suggestion
          ha_cluster_fence_agent_packages }}"
```

tasks/shell_crmsh/check-and-prepare-role-variables.yml
```diff
@@ -0,0 +1,127 @@
...
+  block:
+    - name: Fail if passwords are not specified
+      ansible.builtin.fail:
+        msg: "{{ item }} must be specified"
+      when:
+        - lookup("vars", item, default="") | string | length < 1
+        - ha_cluster_cluster_present | bool
+      loop:
+        - ha_cluster_hacluster_password
+      run_once: true
```
richm, line 13: If there is any chance this task will display the cleartext password when using a verbose log level (e.g. `ansible-playbook -vv ...`), then you should add `no_log: true` to this task

tasks/shell_crmsh/check-and-prepare-role-variables.yml
```diff
@@ -0,0 +1,127 @@
...
+
+    - name: Fail if nodes do not have the same number of SBD devices specified
+      ansible.builtin.fail:
+        msg: All nodes must have the same number of SBD devices specified
+      when:
+        - ha_cluster_cluster_present
```
richm, line 19:
```suggestion
        - ha_cluster_cluster_present | bool
```

tasks/shell_crmsh/check-and-prepare-role-variables.yml
```diff
@@ -0,0 +1,127 @@
...
+    # Running a qnetd on a cluster node does't make sense, fencing would make
+    # the qnetd unavailable, even if temporarily.
+    - name: Fail if configuring qnetd on a cluster node
+      ansible.builtin.fail:
+        msg: >
+          Qnetd cannot be configured on a cluster node -
+          'ha_cluster_cluster_present' and 'ha_cluster_qnetd.present' cannot
+          be both set to true
+      when:
+        - ha_cluster_cluster_present
```
richm, line 37:
```suggestion
        - ha_cluster_cluster_present | bool
```

tasks/shell_crmsh/cluster-destroy-crm.yml
```diff
@@ -0,0 +1,67 @@
+# SPDX-License-Identifier: MIT
+---
+- name: Set fact list of configuration files
+  ansible.builtin.set_fact:
+    __ha_cluster_config_files:
```
richm, line 5: since you don't use __ha_cluster_config_files anywhere else except for the `stat` you can define this variable as a `vars` on the `stat` task (or simply eliminate the variable altogether, unless you intend to add more files or use this elsewhere)

tasks/shell_crmsh/cluster-destroy-crm.yml
```diff
@@ -0,0 +1,67 @@
...
+      - /etc/corosync/corosync.conf
+      - /var/lib/pacemaker/cib/cib.xml
+
+- name: Get stat of cluster configuration files
+  ansible.builtin.stat:
+    path: "{{ item }}"
+  loop: "{{ __ha_cluster_config_files }}"
```
richm, line 12:
```suggestion
  loop:
      - /etc/corosync/corosync.conf
      - /var/lib/pacemaker/cib/cib.xml
```

tasks/shell_crmsh/cluster-destroy-crm.yml
richm: In general, it is better to avoid using `set_fact` as it "pollutes" the global namespace, while using `vars` can be scoped to a context

tasks/shell_crmsh/cluster-setup-keys.yml
```diff
@@ -0,0 +1,35 @@
...
+    preshared_key_length: 256
+
+- name: Distribute corosync authkey
+  ansible.builtin.copy:
+    content: "{{ __ha_cluster_some_preshared_key | b64decode }}"
+    dest: /etc/corosync/authkey
+    owner: root
+    group: root
+    mode: '0400'
+  register: __ha_cluster_distribute_corosync_authkey
```
richm, line 18: not sure if `content` is marked as a `no_log` parameter in the `copy` module, but if you see it being logged with `ansible-playbook -vv`, you should mark this task with `no_log: true` - same with `Distribute pacemaker authkey` below

tasks/shell_crmsh/cluster-start-and-reload.yml
```diff
@@ -0,0 +1,107 @@
...
+  delay: 10
+  timeout: 120
+
+- name: Output current cluster status
+  ansible.builtin.debug:
+    var: __ha_cluster_crm_output_online.stdout_lines
+
+- name: Filter out crm status output
+  ansible.builtin.set_fact:
+    __ha_cluster_crm_status_online: "{{ __ha_cluster_crm_output_online.stdout_lines | select('search', 'Online:')
```
richm, line 98:
```suggestion
    __ha_cluster_crm_status_online: "{{ __ha_cluster_crm_output_online.stdout_lines |
      select('search', 'Online:') | list }}"
```

tasks/shell_crmsh/cluster-start-and-reload.yml
```diff
@@ -0,0 +1,107 @@
...
+     | list }}"
```
richm, line 99:
```suggestion
```

tasks/shell_crmsh/create-and-push-cib.yml
```diff
@@ -0,0 +1,266 @@
...
+# Create/Replace shadow configuration with new based on current cluster.
+- name: Create new shadow crm configuration with force
+  ansible.builtin.command:
+    cmd: crm configure cib new {{ __ha_cluster_crm_shadow }} --force
+  check_mode: false
+  changed_when: not ansible_check_mode
+
+# Build the new CIB
+- name: Build the new CIB
+  block:
```
richm, line 99: why use a `block` here?

tasks/shell_crmsh/crm-cib-constraint-location.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: Delete present location constraint {{ __ha_cluster_constraint_id }}
+  ansible.builtin.command:
+    cmd: crm -c {{ __ha_cluster_crm_shadow }} configure delete {{ __ha_cluster_constraint_id }}
+  when: __ha_cluster_constraint_status.rc == 0
+  check_mode: false
+  changed_when: not ansible_check_mode
+
+- name: Configure location constraint {{ __ha_cluster_constraint_id }}
+  ansible.builtin.command:
+    cmd: |
```
richm, line 28: note that for complex templates, you can define these in a templates/ file and use the `template` lookup to convert the template to a string that you can pass to `command`, `copy.content`, etc.
this is fine, but fyi for future - https://docs.ansible.com/ansible/latest/collections/ansible/builtin/template_lookup.html#ansible-collections-ansible-builtin-template-lookup

vars/RedHat.yml
```diff
@@ -41,3 +41,10 @@ __ha_cluster_services:
   - corosync
   - corosync-qdevice
   - pacemaker
+
+# fence agent list moved from defaults
+ha_cluster_fence_agent_packages: "{{
```
richm, line 46: @tomjelinek was the reason that this was originally in defaults/main.yml because users might want to set this?  If so, we'll have to figure out how to do this in defaults/main.yml, and figure out how to set it to a default value defined in vars/RedHat.yml

tasks/shell_crmsh/cluster-destroy-crm.yml
sean-freeman: Makes sense to me, cannot see `__ha_cluster_config_files` private variable used elsewhere and the files list is unlikely to change in future PCMK versions or OS versions?

Proposal below to use directly in the loop is most condensed method, or the alternative to add a task-level `vars:` underneath the loop.

tasks/shell_crmsh/create-and-push-cib.yml
sean-freeman: I assume to maintain structure equality between the `shell_pcs` and `shell_crmsh` subdirectories and code?

Ref: https://github.com/linux-system-roles/ha_cluster/blob/1.13.5/tasks/shell_pcs/create-and-push-cib.yml#L77-L87

richm: the pcs code uses a block because it defines `vars` that are scoped to the block:
```yaml
- name: Build the new CIB
  vars:
    __ha_cluster_pcs_cli_role:
      promoted: "{{
        ('pcmk.cib.roles.promoted-unpromoted' in __ha_cluster_pcs_capabilities)
        | ternary('Promoted', 'Master') }}"  # wokeignore:rule=master
      unpromoted: "{{
        ('pcmk.cib.roles.promoted-unpromoted' in __ha_cluster_pcs_capabilities)
        | ternary('Unpromoted', 'Slave') }}"  # wokeignore:rule=slave
  block:
```
you only need to use a `block` if 1) you are defining `vars` scoped to the block 2) you have several tasks you want to conditionally execute using a `when` and you don't want to put the `when` on all of those tasks 3) you want to catch errors using `rescue` and run cleanup using `always`

tasks/shell_crmsh/crm-cib-constraint-location.yml
sean-freeman: Looks like another structure equality between the shell_pcs and shell_crmsh subdirectories and code?

@tomjelinek should weigh in here too, as this might be tech debt to address across the whole Ansible Role dating back to Ansible Core 2.9 compatibility (when template lookup was very different) ?

Ref: https://github.com/linux-system-roles/ha_cluster/blob/1.13.5/tasks/shell_pcs/pcs-cib-constraint-location.yml#L3-L6

vars/RedHat.yml
sean-freeman: Would the following in `defaults/main.yml` work?

```yaml
var1_packages_list: "{{ var2_packages_dict[ansible_os_family] }}"

var2_packages_dict:
  RedHat:
    - value1
    - value2
    - value3
  Suse:
    - value1
    - value2
    - value3
```

tasks/shell_crmsh/create-and-push-cib.yml
sean-freeman: As the code is functional, and to try to maintain the structure equality - can this be amendment be skipped? This is then clearer for future code review / maintenance, and can append with the block-scoped `vars` if needed in future? 

Minor alteration to remove the block and append in future if needed, so not something that needs debate. If you prefer clean, the block can be removed.

tasks/shell_crmsh/crm-cib-constraint-location.yml
richm: Sure - just FYI - most devs don't know about the `template` lookup, and think they can only use template files with the `template` module.

vars/RedHat.yml
richm: Something like that, yes

tasks/shell_crmsh/check-and-prepare-role-variables.yml
marcelmamula: I will add it, but pcs file also does not have it since it is one of tasks that was copied  ... do you want me to add it to pcs as well?

https://github.com/linux-system-roles/ha_cluster/blob/fcd3461ec792a9b5de3a01a6f423d8d5a6942833/tasks/shell_pcs/check-and-prepare-role-variables.yml#L15C1-L25C21

tasks/shell_crmsh/cluster-destroy-crm.yml
marcelmamula: You are correct, i added it just in case there will be more files in future. I am changing it as mentioned below.

tasks/shell_crmsh/cluster-destroy-crm.yml
marcelmamula: Done

tasks/shell_crmsh/cluster-setup-keys.yml
marcelmamula: Fixed, but pcs code also does not have it. Do you want to include it as well?

tasks/shell_crmsh/cluster-start-and-reload.yml
marcelmamula: Fixed

tasks/shell_crmsh/crm-cib-constraint-location.yml
marcelmamula: My intent was structure equality as Sean pointed out, plus I wanted avoid creation of temporary files with template module, so I opted to follow same as you did with pcs inline jinja2.

It was not easy, but current j2 code is generating crmsh compliant commands and creates resources same as pcs.

vars/RedHat.yml
tomjelinek: > @tomjelinek was the reason that this was originally in defaults/main.yml because users might want to set this?

Yes, it was. See https://github.com/linux-system-roles/ha_cluster#ha_cluster_fence_agent_packages

templates/corosync.j2
```diff
@@ -0,0 +1,49 @@
...
+        clear_node_high_bit: {{ ha_cluster_corosync_params.totem.clear_node_high_bit | default("yes") }}
+        rrp_mode: {{ ha_cluster_corosync_params.totem.rrp_mode | default("passive") }}
+
+        interface {
+                ringnumber: 0
+                bindnetaddr: {{ ansible_facts['eth0']['ipv4']['address'] }}
+                mcastport: 5405
+                ttl: 1
+        }
+        transport: udpu
```
tomjelinek, line 18: If Suse doesn't support knet, it should be mentioned in the documentation, I guess. Is Suse still running corosync 2.x? It looks like that branch is no longer supported: https://github.com/corosync/corosync/wiki/Releases

Also, it would be nice if `tasks/shell_crmsh/check-and-prepare-role-variables.yml` failed when corosync options not working on Suse are specified.

templates/corosync.j2
```diff
@@ -0,0 +1,49 @@
...
+                subsys: QUORUM
+                debug: off
+        }
+}
+nodelist {
+{% for host in ansible_play_batch %}
+       node {
+               ring0_addr: {{ hostvars[host]['ansible_facts']['eth0']['ipv4']['address'] }}
+               {% if hostvars[host]['ansible_facts']['eth1']['ipv4']['address'] is defined %}
+               ring1_addr: {{ hostvars[host]['ansible_facts']['eth1']['ipv4']['address'] }}
```
tomjelinek, line 37: No IPv6 support? What if the network interfaces have other names than eth0, eth1 (I'm not sure if that can happen on Suse).

templates/corosync.j2
tomjelinek, line 1: Should we create pcs and crmsh directories in templates and move this file into templates/crmsh? That would make it clear that it's not used by pcs part of the role. Likewise, pcs-only templates would be moved to templates/pcs.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
tomjelinek, line 1: Major part of this file is the same for both pcs and crmsh. How about splitting it to a common file and shell-specific files instead of duplicating the parts which are the same?

vars/Suse.yml
```diff
@@ -1,3 +1,40 @@
...
+  - 'ha-cluster-bootstrap'
+  - 'pacemaker'
+  - 'patterns-ha-ha_sles'
+  - 'resource-agents'
+  - 'cluster-glue'
+  - 'socat'
+  - 'libxml2-tools'
+  - 'ClusterTools2' # not pre-installed in SLES for SAP
+  - 'SAPHanaSR'
+  - 'rsyslog'
```
tomjelinek, line 21: The essential packages variable is supposed to hold only a minimal set of packages which are required by the role to run. For example, when configuring a qnetd node, you don't need pacemaker, corosync, fence-agents, resource-agents, etc. installed there. Those only get installed when `ha_cluster_cluster_present` is `true`.

tomjelinek (changes requested):
I'm getting a lot of errors from linters in CI. Please, fix those:

yamllint:
```
./vars/Suse.yml
  19:21     warning  too few spaces before comment  (comments)
  23:81     error    line too long (82 > 80 characters)  (line-length)
./tasks/main.yml
  95:81     error    line too long (87 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-constraint-location.yml
  7:81      error    line too long (99 > 80 characters)  (line-length)
  8:81      error    line too long (104 > 80 characters)  (line-length)
  13:81     error    line too long (93 > 80 characters)  (line-length)
  18:81     error    line too long (85 > 80 characters)  (line-length)
  21:81     error    line too long (95 > 80 characters)  (line-length)
  29:81     error    line too long (94 > 80 characters)  (line-length)
./tasks/shell_crmsh/create-and-push-cib.yml
  237:81    error    line too long (81 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-constraint-ticket.yml
  12:81     error    line too long (93 > 80 characters)  (line-length)
  17:81     error    line too long (85 > 80 characters)  (line-length)
  20:81     error    line too long (95 > 80 characters)  (line-length)
  28:81     error    line too long (96 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-constraint-set.yml
  3:88      warning  too few spaces before comment  (comments)
  3:81      error    line too long (108 > 80 characters)  (line-length)
  12:81     error    line too long (93 > 80 characters)  (line-length)
  17:81     error    line too long (85 > 80 characters)  (line-length)
  20:81     error    line too long (95 > 80 characters)  (line-length)
  25:89     warning  too few spaces before comment  (comments)
  25:81     error    line too long (109 > 80 characters)  (line-length)
  33:81     error    line too long (81 > 80 characters)  (line-length)
  39:81     error    line too long (82 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-resource-primitive.yml
  11:81     error    line too long (83 > 80 characters)  (line-length)
  20:28     warning  too few spaces before comment  (comments)
./tasks/shell_crmsh/crm-cib-stonith-level.yml
  12:81     error    line too long (88 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cluster-properties.yml
  6:81      error    line too long (110 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-constraint-order.yml
  12:81     error    line too long (93 > 80 characters)  (line-length)
  17:81     error    line too long (85 > 80 characters)  (line-length)
  20:81     error    line too long (95 > 80 characters)  (line-length)
  28:81     error    line too long (99 > 80 characters)  (line-length)
  29:81     error    line too long (95 > 80 characters)  (line-length)
  30:81     error    line too long (93 > 80 characters)  (line-length)
./tasks/shell_crmsh/sbd.yml
  15:81     error    line too long (82 > 80 characters)  (line-length)
  16:81     error    line too long (92 > 80 characters)  (line-length)
  17:81     error    line too long (81 > 80 characters)  (line-length)
  18:81     error    line too long (93 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-resource-group.yml
  6:81      error    line too long (84 > 80 characters)  (line-length)
  11:81     error    line too long (83 > 80 characters)  (line-length)
  14:81     error    line too long (86 > 80 characters)  (line-length)
./tasks/shell_crmsh/cluster-start-and-reload.yml
  64:81     error    line too long (85 > 80 characters)  (line-length)
  98:81     error    line too long (85 > 80 characters)  (line-length)
  103:81    error    line too long (86 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-resource-clone.yml
  3:81      error    line too long (113 > 80 characters)  (line-length)
  6:81      error    line too long (106 > 80 characters)  (line-length)
  11:81     error    line too long (91 > 80 characters)  (line-length)
  16:81     error    line too long (83 > 80 characters)  (line-length)
  19:81     error    line too long (93 > 80 characters)  (line-length)
  27:81     error    line too long (89 > 80 characters)  (line-length)
./tasks/shell_crmsh/crm-cib-constraint-colocation.yml
  3:81      error    line too long (81 > 80 characters)  (line-length)
  12:81     error    line too long (93 > 80 characters)  (line-length)
  17:81     error    line too long (85 > 80 characters)  (line-length)
  20:81     error    line too long (95 > 80 characters)  (line-length)
  28:81     error    line too long (96 > 80 characters)  (line-length)
  34:81     error    line too long (97 > 80 characters)  (line-length)
```

ansible-lint:
```
yaml[line-length]: Line too long (87 > 80 characters)
tasks/main.yml:95
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/cluster-start-and-reload.yml:64
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/cluster-start-and-reload.yml:98
yaml[line-length]: Line too long (86 > 80 characters)
tasks/shell_crmsh/cluster-start-and-reload.yml:103
yaml[line-length]: Line too long (81 > 80 characters)
tasks/shell_crmsh/create-and-push-cib.yml:237
yaml[line-length]: Line too long (81 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-colocation.yml:3
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-colocation.yml:12
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-colocation.yml:17
yaml[line-length]: Line too long (95 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-colocation.yml:20
yaml[line-length]: Line too long (96 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-colocation.yml:28
yaml[line-length]: Line too long (97 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-colocation.yml:34
yaml[line-length]: Line too long (99 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-location.yml:7
yaml[line-length]: Line too long (104 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-location.yml:8
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-location.yml:13
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-location.yml:18
yaml[line-length]: Line too long (95 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-location.yml:21
yaml[line-length]: Line too long (94 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-location.yml:29
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-order.yml:12
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-order.yml:17
yaml[line-length]: Line too long (95 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-order.yml:20
yaml[line-length]: Line too long (99 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-order.yml:28
yaml[line-length]: Line too long (95 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-order.yml:29
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-order.yml:30
yaml[line-length]: Line too long (108 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:3
yaml[comments]: Too few spaces before comment
tasks/shell_crmsh/crm-cib-constraint-set.yml:3
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:12
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:17
yaml[line-length]: Line too long (95 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:20
yaml[line-length]: Line too long (109 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:25
yaml[comments]: Too few spaces before comment
tasks/shell_crmsh/crm-cib-constraint-set.yml:25
yaml[line-length]: Line too long (81 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:33
yaml[line-length]: Line too long (82 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-set.yml:39
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-ticket.yml:12
yaml[line-length]: Line too long (85 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-ticket.yml:17
yaml[line-length]: Line too long (95 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-ticket.yml:20
yaml[line-length]: Line too long (96 > 80 characters)
tasks/shell_crmsh/crm-cib-constraint-ticket.yml:28
yaml[line-length]: Line too long (113 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-clone.yml:3
yaml[line-length]: Line too long (106 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-clone.yml:6
yaml[line-length]: Line too long (91 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-clone.yml:11
yaml[line-length]: Line too long (83 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-clone.yml:16
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-clone.yml:19
yaml[line-length]: Line too long (89 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-clone.yml:27
yaml[line-length]: Line too long (84 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-group.yml:6
yaml[line-length]: Line too long (83 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-group.yml:11
yaml[line-length]: Line too long (86 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-group.yml:14
yaml[line-length]: Line too long (83 > 80 characters)
tasks/shell_crmsh/crm-cib-resource-primitive.yml:11
yaml[comments]: Too few spaces before comment
tasks/shell_crmsh/crm-cib-resource-primitive.yml:20
yaml[line-length]: Line too long (88 > 80 characters)
tasks/shell_crmsh/crm-cib-stonith-level.yml:12
yaml[line-length]: Line too long (110 > 80 characters)
tasks/shell_crmsh/crm-cluster-properties.yml:6
yaml[line-length]: Line too long (82 > 80 characters)
tasks/shell_crmsh/sbd.yml:15
yaml[line-length]: Line too long (92 > 80 characters)
tasks/shell_crmsh/sbd.yml:16
yaml[line-length]: Line too long (81 > 80 characters)
tasks/shell_crmsh/sbd.yml:17
yaml[line-length]: Line too long (93 > 80 characters)
tasks/shell_crmsh/sbd.yml:18
yaml[comments]: Too few spaces before comment
vars/Suse.yml:19
yaml[line-length]: Line too long (82 > 80 characters)
vars/Suse.yml:23
```

tasks/shell_crmsh/create-and-push-cib.yml
tomjelinek: Sorry, I don't follow. Why would it be needed to put the cluster into the maintenance mode?

When using crmsh manually from command line, is the user required to put a cluster into the maintenance mode before doing any changes to the cluster configuration? I haven't heard that pacemaker would require that. Corosync doesn't have a maintenance mode at all, how would you change its configuration then?

tasks/shell_pcs/cluster-configure.yml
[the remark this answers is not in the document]
marcelmamula: Which name would fit there so it is not shell specific?

templates/corosync.j2
```diff
@@ -0,0 +1,49 @@
+totem {
```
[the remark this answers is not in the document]
marcelmamula: Thank you, added

templates/corosync.j2
marcelmamula: I can do that by creating crmsh folder, or we can just prepend them? crmsh-corosync.j2 There are not that many files yet

templates/corosync.j2
marcelmamula: This is initial setup for Suse so it does not account for all potential possibilities. I think they can be added subsequently, along with updated readme to account for corosync variable.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
marcelmamula: I agree. Would it be best to do it after your currently open PR for same file to avoid conflicts?

vars/Suse.yml
marcelmamula: I agree, I will move them under __ha_cluster_fullstack_node_packages instead.

tasks/main.yml
tomjelinek: When the role is using pcs, all CIB changes are done to a temporary file, and there is no harm in doing so on all nodes. Once the CIB is built in the file, it is being pushed to pacemaker on one node only: `run_once: true` is set for "Push CIB diff to the cluster if it has any changes" task in `create-and-push-cib.yml`. For crmsh, pushing the CIB to pacemaker is limited to one node in the same way.

I don't see any benefit in including `create-and-push-cib.yml` on one node only, at least for pcs. CIB changes are done on only one node anyway by the mechanism described above.

marcelmamula: You are correct, but it also means you do them on non-primary node(s) and delete anyway, which could just be avoided with run_once.

tasks/shell_crmsh/create-and-push-cib.yml
marcelmamula: cibadmin --query is sensitive on Time and cluster status, which means that I was able to reproduce 20-50% chance for cibadmin patch to result in error with outdated version, just because cib build workflow was working on already outdated cib export.

This step freezes cluster, waits for it to reflect freeze and then gets accurate cib, which will be accepted in the end.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
tomjelinek: This doesn't print the password even with `-vv` for me...

tasks/shell_crmsh/cluster-setup-keys.yml
tomjelinek: It seems to be `no_log: true`: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/copy.py#L537

tasks/shell_crmsh/check-and-prepare-role-variables.yml
sean-freeman: Seems sensible to wait until after this initial compatibility PR is added

tasks/shell_crmsh/check-and-prepare-role-variables.yml
marcelmamula: thank you for checking, I will keep it in crmsh just in case.

templates/corosync.j2
marcelmamula: Current version is 2.4.6 and 3.1.x is planned for future.

tasks/shell_crmsh/crm-cib-constraint-location.yml
tomjelinek: I didn't even know it was possible to use template files in this way. Still, I like the inline templates more for this purpose, it reduces the number of files in the role and thus it makes maintenance a bit easier. Also, if the command templates were separate files, one could think that they are used as a basis for files produced by the role, which is not true.

tasks/shell_pcs/cluster-configure.yml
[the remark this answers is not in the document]
tomjelinek: How about configure-shell.yml? It can be configure-pcs.yml and configure-crmsh.yml as well, since the shell name variable is used for the directory anyway.

templates/corosync.j2
tomjelinek: OK, that makes sense. Start with the basics, add complexities and features later. :+1:

templates/corosync.j2
tomjelinek: Prefixing works for me as well. It may be even a better options, as there are not many templates.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
tomjelinek: Yes, works for me.

tasks/shell_pcs/cluster-configure.yml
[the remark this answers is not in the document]
marcelmamula: I think we can go with configure-shell.yml as you mentioned:
shell_{{ ha_cluster_pacemaker_shell }}/configure-shell.yml

ha_cluster_pacemaker_shell is crmsh and files under crmsh contain name crm instead of crmsh.

templates/corosync.j2
marcelmamula: Good, i will update it in upcoming commit.

tasks/shell_pcs/cluster-configure.yml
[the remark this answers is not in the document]
tomjelinek: Ah, ok. configure-shell it is, then.

tasks/shell_crmsh/create-and-push-cib.yml
tomjelinek: This should be prevented by specifying `--no-version` option for `crm_diff`, which both pcs and crmsh parts do. I haven't run into the outdated cib error even once with the role and it wasn't reported by anyone, as far as I know, either.

tasks/main.yml
sean-freeman: Given it is all automated, it does not matter if some steps are repeated/duplicated and then not used. If there is a difference here between `pcs` and `crmsh` shell's, I don't see an impact.

However logically speaking - I conversely don't see any benefit for `pcs` having the CIB file on 2..n nodes if it's only done on 1 node anyway. Seems pointless to create the file that doesn't get used. Not worth debating.

@tomjelinek can you please confirm how you want to proceed?  Leave this `run_once` when using `crmsh` PCMK Shell ?

tasks/shell_crmsh/crm-cib-constraint-location.yml
richm: > I didn't even know it was possible to use template files in this way. Still, I like the inline templates more for this purpose, it reduces the number of files in the role and thus it makes maintenance a bit easier.

Ok.

> Also, if the command templates were separate files, one could think that they are used as a basis for files produced by the role, which is not true.

Ok.

vars/RedHat.yml
richm: > Can we keep ha_cluster_fence_agent_packages as user input defined in defaults as [] and change actual variable to private **__ha_cluster_fence_agent_packages** which is defined from os_family vars depending if ha_cluster_fence_agent_packages | length > 0, otherwise use user input from ha_cluster_fence_agent_packages ?

yes, sounds good

tasks/shell_crmsh/cluster-setup-keys.yml
sean-freeman: @tomjelinek confirms `no_log: true` is default for `copy` module, so is this change really needed?

richm:
If you need help with line wrapping, see https://linux-system-roles.github.io/documentation/howto/working-with-ansible-jinja-code-and-filters.html "How to solve some common ansible-lint issues"

marcelmamula:
> If you need help with line wrapping, see https://linux-system-roles.github.io/documentation/howto/working-with-ansible-jinja-code-and-filters.html "How to solve some common ansible-lint issues"

My ansible lint is not showing line length warnings, but I am updating them now based on Tomas's list.

tasks/shell_crmsh/cluster-setup-keys.yml
richm: > It seems to be `no_log: true`: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/copy.py#L537

Ok - as long as no cleartext passwords/keys are logged with `-vv`, you don't have to do anything

tasks/shell_crmsh/crm-cib-constraint-location.yml
sean-freeman: Decision made, @richm please mark as resolved if you agree.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
richm: > Same comment as above, pcs code does not have it either.

hmm - ok - I guess ansible-lint doesn't check for this anymore?

richm:
> > If you need help with line wrapping, see https://linux-system-roles.github.io/documentation/howto/working-with-ansible-jinja-code-and-filters.html "How to solve some common ansible-lint issues"
> 
> My ansible lint is not showing line length warnings, but I am updating them now based on Tomas's list.

How are you running ansible-lint?

tasks/main.yml
tomjelinek: Fair point. Right now, I don't see a strict requirement for running pcs commands building cib on all nodes. Let's make this `run_once`, then.

tasks/shell_crmsh/create-and-push-cib.yml
sean-freeman: > Sorry, I don't follow. Why would it be needed to put the cluster into the maintenance mode?

@tomjelinek My original comment was focused on the general principle/usage of any "maintenance mode" as an industry-wide best practice for any in-place software re-configuration/update. Seems logical if the maintenance mode exists, we should use it as intended. Clearly the Ansible Role has worked (very well) to date, without needing this. I was only highlighting whether it would be good to include in `pcs` PCMK Shell execution flow, as it seems logical.

.... I guess the question becomes, `"Why do we NOT use 'pcs' PCMK Shell in maintenance mode before making changes?"`

---

Although from the continuing discussion with @marcelmamula , it looks like maintenance mode is a firm requirement when using `crmsh` PCMK Shell.

richm:
This repo CI uses the .ansible-lint in the ha_cluster root directory.  https://github.com/linux-system-roles/ha_cluster/blob/main/.ansible-lint
```yaml
skip_list:
  - fqcn-builtins
  - var-naming[no-role-prefix]
  - sanity[cannot-ignore]  # wokeignore:rule=sanity
```
The reason why we don't see line length errors in the github action CI is because the latest ansible-lint github action only works on collections - so we have to convert the role to collection format before running ansible-lint - https://github.com/linux-system-roles/ha_cluster/blob/main/.github/workflows/ansible-lint.yml#L37 - when we convert to collection format, we cannot preserve line wrapping (due to ruamel roundtrip), so we disable line-length checking.

tasks/main.yml
sean-freeman: Agreed 🙂 The `run_once` for the building CIB file on node 1 only will be performed for `crmsh` PCMK Shell, and current logic for building CIB file on all nodes will be performed for `pcs` PCMK Shell. Marking this resolved.

tomjelinek:
> This repo CI uses the .ansible-lint in the ha_cluster root directory. https://github.com/linux-system-roles/ha_cluster/blob/main/.ansible-lint
> 
> ```yaml
> skip_list:
>   - fqcn-builtins
>   - var-naming[no-role-prefix]
>   - sanity[cannot-ignore]  # wokeignore:rule=sanity
> ```
> 
> The reason why we don't see line length errors in the github action CI is because the latest ansible-lint github action only works on collections - so we have to convert the role to collection format before running ansible-lint - https://github.com/linux-system-roles/ha_cluster/blob/main/.github/workflows/ansible-lint.yml#L37 - when we convert to collection format, we cannot preserve line wrapping (due to ruamel roundtrip), so we disable line-length checking.

@richm Does that mean we don't care about line length anymore? Should I disable those checks in my tests? I still run ansible-lint and yamllint on the original source files.

vars/RedHat.yml
marcelmamula: change in coming commit, bit changed but simplified. User input retained while os_family vars will kick in if it is empty from defaults.

```
# fence agent list for os_family
__ha_cluster_fence_agent_packages: "{{
    ['fence-agents-all']
    +
    (['fence-virt'] if ansible_architecture == 'x86_64' else [])
  }}"

# fence agent assignment based on user input
ha_cluster_fence_agent_packages: "{{ ha_cluster_fence_agent_packages if
 ha_cluster_fence_agent_packages | length > 0
 else __ha_cluster_fence_agent_packages }}"
```

richm:
> > This repo CI uses the .ansible-lint in the ha_cluster root directory. https://github.com/linux-system-roles/ha_cluster/blob/main/.ansible-lint
> > ```yaml
> > skip_list:
> >   - fqcn-builtins
> >   - var-naming[no-role-prefix]
> >   - sanity[cannot-ignore]  # wokeignore:rule=sanity
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The reason why we don't see line length errors in the github action CI is because the latest ansible-lint github action only works on collections - so we have to convert the role to collection format before running ansible-lint - https://github.com/linux-system-roles/ha_cluster/blob/main/.github/workflows/ansible-lint.yml#L37 - when we convert to collection format, we cannot preserve line wrapping (due to ruamel roundtrip), so we disable line-length checking.
> 
> @richm Does that mean we don't care about line length anymore?

I think we should care about line length - not sure if Galaxy role gating checks.  I'm considering putting in a separate github action check to run ansible-lint against the role.

> Should I disable those checks in my tests?

No, I think we should still care about line length.

> I still run ansible-lint and yamllint on the original source files.

ansible-lint now runs yamllint, so you shouldn't need a separate check (and yes, ansible-lint will use .yamllint.yml if you have one).

vars/RedHat.yml
richm: ok

vars/Suse.yml
marcelmamula: I have removed SAPHanaSR, but I am keeping rest in. qnetd setup is not part of this PR so current functionality focuses on cluster, where these packages are required.

Future PR focused on qnet/qdevice should be able to carve out essentials.

richm: > Future PR focused on qnet/qdevice should be able to carve out essentials.

Yes.  Let's focus this PR on the changes required to the common role functionality (code that affects both the pcs and the crmsh implementations), then hopefully future changes will only affect code specific to crmsh and will be easier to merge.

vars/RedHat.yml
marcelmamula: Fixed in current commit

tasks/shell_crmsh/create-and-push-cib.yml
marcelmamula: Maintenance mode was enhanced to cover edge case with is-managed meta attribute

tasks/shell_crmsh/crm-cib-resource-primitive.yml
```diff
@@ -0,0 +1,51 @@
...
+# Delete resource id in Shadow CIB to avoid errors during cibadmin patch.
+- name: Delete present resource primitive {{ resource.id }}
+  ansible.builtin.command:
+    cmd: crm -c {{ __ha_cluster_crm_shadow }} configure delete {{ resource.id }}
+  when: __ha_cluster_resource_status.rc == 0
+  check_mode: false
+  changed_when: not ansible_check_mode
+
+- name: Configure resource primitive {{ resource.id }}
+  ansible.builtin.command: # noqa jinja[spacing]
```
richm, line 20:
```suggestion
  ansible.builtin.command:  # noqa jinja[spacing]
```

richm:
[citest]
just want to make sure there are no regressions in the RedHat os_family pcs tests

defaults/main.yml
```diff
@@ -13,11 +13,7 @@ ha_cluster_start_on_boot: true
 
 ha_cluster_extra_packages: []
 
-ha_cluster_fence_agent_packages: "{{
-    ['fence-agents-all']
-    +
-    (['fence-virt'] if ansible_architecture == 'x86_64' else [])
-  }}"
+ha_cluster_fence_agent_packages: []
```
richm, line 16:
```suggestion
ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages }}"
```

marcelmamula: I don't think that is correct place, because it would expose private variable to potential user inputs.

I have added conditional into main yesterday to do switch, depending if users added something to it.
```
+
      ha_cluster_fence_agent_packages
        if ha_cluster_fence_agent_packages | length > 0
        else __ha_cluster_fence_agent_packages }}"
```

marcelmamula:
I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.

tomjelinek:
> I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.

Tests are failing with the same errors on my systems as well. This needs to be investigated.

marcelmamula:
> > I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.
> 
> Tests are failing with the same errors on my systems as well. This needs to be investigated.

I tried running this PR and existing collection (without PR) on CentOS9, but it fails in different place, which seems as wrong order of steps in pcs. So i cannot re-test this with my current setup. **Let me know if you identify something., since it seems you have more variables defined to overcome error below.**

Issue I have in mine are from pcs attempting to do property change (empty value) on nonexistent CIB
`fatal: [centos0]: FAILED! => {"changed": true, "cmd": ["pcs", "--force", "-f", "/var/lib/pacemaker/cib/cib.xml", "--", "property", "set", "stonith-watchdog-timeout=0"], "delta": "0:00:00.992301", "end": "2024-02-09 09:22:47.144951", "msg": "non-zero return code", "rc": 1, "start": "2024-02-09 09:22:46.152650", "stderr": "Error: unable to get cib", "stderr_lines": ["Error: unable to get cib"], "stdout": "", "stdout_lines": []}`

tomjelinek:
> I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.

The firewall failures are caused by not enabling port 1229/tcp. This port is required for fence-virt. You broke this by changes related to `ha_cluster_fence_agent_packages` variable. Check `tasks/firewall.yml` - `ha_cluster_fence_agent_packages` is empty there, thus fence-virt port is not enabled. The `ha_cluster_fence_agent_packages` variable gets populated by `__ha_cluster_fence_agent_packages` in `tasks/main.yml` after that. That's the root cause.

marcelmamula:
> > I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.
> 
> The firewall failures are caused by not enabling port 1229/tcp. This port is required for fence-virt. This got broken by changes related to `ha_cluster_fence_agent_packages` variable. Check `tasks/firewall.yml` - `ha_cluster_fence_agent_packages` is empty there, thus fence-virt port is not enabled. The `ha_cluster_fence_agent_packages` variable gets populated by `__ha_cluster_fence_agent_packages` in `tasks/main.yml` after that. That's the root cause.

I see, I have tried to add it to include_vars/os_family yesterday but it resulted in critical nesting, so I moved it to main.yml into "Install cluster packages" as 
```
+
          ha_cluster_fence_agent_packages
            if ha_cluster_fence_agent_packages | length > 0
            else __ha_cluster_fence_agent_packages }}"
```

But there are other places that use it before that - firewall and selinux. Where would be best place to do this assignment then if include_vars cannot be used with same syntax?

Nesting came from this in include_vars
```
ha_cluster_fence_agent_packages: "{{ ha_cluster_fence_agent_packages if
 ha_cluster_fence_agent_packages | length > 0
 else __ha_cluster_fence_agent_packages }}"
```

marcelmamula:
> > > I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.
> > 
> > 
> > The firewall failures are caused by not enabling port 1229/tcp. This port is required for fence-virt. This got broken by changes related to `ha_cluster_fence_agent_packages` variable. Check `tasks/firewall.yml` - `ha_cluster_fence_agent_packages` is empty there, thus fence-virt port is not enabled. The `ha_cluster_fence_agent_packages` variable gets populated by `__ha_cluster_fence_agent_packages` in `tasks/main.yml` after that. That's the root cause.
> 
> I see, I have tried to add it to include_vars/os_family yesterday but it resulted in critical nesting, so I moved it to main.yml into "Install cluster packages" as
> 
> ```
> +
>           ha_cluster_fence_agent_packages
>             if ha_cluster_fence_agent_packages | length > 0
>             else __ha_cluster_fence_agent_packages }}"
> ```
> 
> But there are other places that use it before that - firewall and selinux. Where would be best place to do this assignment then if include_vars cannot be used with same syntax?
> 
> Nesting came from this in include_vars
> 
> ```
> ha_cluster_fence_agent_packages: "{{ ha_cluster_fence_agent_packages if
>  ha_cluster_fence_agent_packages | length > 0
>  else __ha_cluster_fence_agent_packages }}"
> ```

@tomjelinek We can simplify it with pre-step in main.yml which avoids nesting and still leaves user input as option.
```
- name: Set ha_cluster_fence_agent_packages fact
  ansible.builtin.set_fact:
    ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages }}"
  when: ha_cluster_fence_agent_packages | length == 0
```

There is also second approach, to always add os_family fence agents on top of user input with | unique, which would be more manageable.

marcelmamula:
@tomjelinek I have updated code for 2nd approach and it works well by combining both and then running unique on top.
Let me know if you are fine with this approach and I will commit it for CI testing.

selinux:
```
(
            'fence-virt' in ha_cluster_fence_agent_packages
            or
            'fence-virt' in __ha_cluster_fence_agent_packages
            or
            'fence-virt' in ha_cluster_extra_packages
            or
            'fence-agents-all' in ha_cluster_fence_agent_packages
            or
            'fence-agents-all' in __ha_cluster_fence_agent_packages
            or
            'fence-agents-all' in ha_cluster_extra_packages
          )
```

firewall
```
    __use_fence_fw_port: "{{ __arch == 'x86_64' and
      ('fence-virt' in ha_cluster_fence_agent_packages
       or 'fence-virt' in __ha_cluster_fence_agent_packages
       or 'fence-virt' in ha_cluster_extra_packages
       or 'fence-agents-all' in ha_cluster_fence_agent_packages
       or 'fence-agents-all' in __ha_cluster_fence_agent_packages
       or 'fence-agents-all' in ha_cluster_extra_packages) }}"
```

main:
```
    - name: Install cluster packages
      package:
        name: "{{(
          __ha_cluster_fullstack_node_packages
          +
          __ha_cluster_qdevice_in_use |
            ternary(__ha_cluster_qdevice_node_packages, [])
          +
          ha_cluster_sbd_enabled | ternary(__ha_cluster_sbd_packages, [])
          +
          ha_cluster_fence_agent_packages
          +
          __ha_cluster_fence_agent_packages ) | unique }}"
        state: present
        use: "{{ (__ha_cluster_is_ostree | d(false)) |
                 ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
```

tasks/main.yml
```diff
@@ -58,7 +58,8 @@
           ha_cluster_sbd_enabled | ternary(__ha_cluster_sbd_packages, [])
           +
           ha_cluster_fence_agent_packages
-        }}"
+            if ha_cluster_fence_agent_packages | length > 0
+            else __ha_cluster_fence_agent_packages }}"
```
tomjelinek, line 62: This `if ... else` influences previous expressions as well, not just `ha_cluster_fence_agent_packages`. It is causing corosync-qdevice not to be installed depending on the value of `ha_cluster_fence_agent_packages`. Either use `ternary` here or brackets.

tomjelinek:
> @tomjelinek I have updated code for 2nd approach and it works well by combining both and then running unique on top.

@marcelmamula That is not going to work. Originally, users could redefine `ha_cluster_fence_agent_packages` value to install only those fence-agents they were interested in. If you combine `ha_cluster_fence_agent_packages` and `__ha_cluster_fence_agent_packages` together, then all fence-agents will be installed every time, as `__ha_cluster_fence_agent_packages` installs fence-agents-all.

tomjelinek:
> I can see Workflow test failed on firewall and qdevice, are those CI environments in good shape? PR does not touch firewall or qdevices.

Qdevice tests are failing due to corosync-qdevice package not being installed by the role. This is caused by changes in "Install cluster packages" task in `main.yml`. I explained this in a comment above.

tomjelinek:
@marcelmamula re: `ha_cluster_fence_agent_packages`

I would try to resolve this by setting fact `__ha_cluster_fence_agent_packages` in `check-and-prepare-role-variables.yml` depending on `ha_cluster_fence_agent_packages` and then using `__ha_cluster_fence_agent_packages` instead of `ha_cluster_fence_agent_packages` throughout the role.

marcelmamula:
> @marcelmamula re: `ha_cluster_fence_agent_packages`
> 
> I would try to resolve this by setting fact `__ha_cluster_fence_agent_packages` in `check-and-prepare-role-variables.yml` depending on `ha_cluster_fence_agent_packages` and then using `__ha_cluster_fence_agent_packages` instead of `ha_cluster_fence_agent_packages` throughout the role.

 @tomjelinek I have retested using ternary `ha_cluster_fence_agent_packages: "{{ ha_cluster_fence_agent_packages | length == 0 | ternary(__ha_cluster_fence_agent_packages, ha_cluster_fence_agent_packages) }}"` and it ends with endless nesting like if/else. 

It might be simplest to do set_facts before firewall/selinux block:
```
- name: Set ha_cluster_fence_agent_packages fact
  ansible.builtin.set_fact:
    ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages }}"
  when: ha_cluster_fence_agent_packages | length == 0
```

tomjelinek:
> I have retested using ternary `ha_cluster_fence_agent_packages: "{{ ha_cluster_fence_agent_packages | length == 0 | ternary(__ha_cluster_fence_agent_packages, ha_cluster_fence_agent_packages) }}"` and it ends with endless nesting like if/else.
> 
> It might be simplest to do set_facts before firewall/selinux block:
> 
> ```
> - name: Set ha_cluster_fence_agent_packages fact
>   ansible.builtin.set_fact:
>     ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages }}"
>   when: ha_cluster_fence_agent_packages | length == 0
> ```

This would work, but it modifies user-defined `ha_cluster_fence_agent_packages` variable, and the modification will be in effect even after the role finishes. Users won't expect this to happen and it may cause issues. I would rather set a "private" variable. If there is endless nesting, perhaps we need a new variable name? `__ha_cluster_fence_agent_packages_final: "{{ ha_cluster_fence_agent_packages | length == 0 | ternary(__ha_cluster_fence_agent_packages_default, ha_cluster_fence_agent_packages) }}"`

Also, I would put this task into `check-and-prepare-role-variables.yml`

tasks/shell_crmsh/create-and-push-cib.yml
sean-freeman: There is no action remaining here, only @tomjelinek to respond to posterity

marcelmamula:
@tomjelinek Thank you for suggestion. I have implemented it and retested in rhel, centos and sles4sap and it was working correctly. New commit was pushed for restesting.

sean-freeman:
[citest]
Maybe this triggers it?

defaults/main.yml
richm: > I don't think that is correct place, because it would expose private variable to potential user inputs.

I'm not sure what you mean.  I believe my proposal encapsulates the desired behavior - putting this in `defaults/main.yml`:
```yaml
ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages }}"
```
This allows users to provide their own list of packages for `ha_cluster_fence_agent_packages` (which is the desired behavior - @tomjelinek please correct me if I'm wrong), and if the user does not specify `ha_cluster_fence_agent_packages`, it will be set to the default value `__ha_cluster_fence_agent_packages`.  And, since `__ha_cluster_fence_agent_packages` is defined for different values depending on the platform/version, we get the correct value for `ha_cluster_fence_agent_packages` for all platforms/versions.

> 
> I have added conditional into main yesterday to do switch, depending if users added something to it.
> 
> ```
> +
>       ha_cluster_fence_agent_packages
>         if ha_cluster_fence_agent_packages | length > 0
>         else __ha_cluster_fence_agent_packages }}"
> ```

I think this is not the "Ansible way" to do this.  I believe the correct way to do this is the way I outlined this above.

tasks/shell_crmsh/create-and-push-cib.yml
tomjelinek: As I mentioned, there haven't been any issues caused by the role not utilizing maintenance mode with pcs, to my knowledge. If maintenance mode is required for the role to run correctly with crmsh, then use it, of course. I'll take a look at the scenario @marcelmamula described to see if I can reproduce the issue using pcs and whether using maintenance mode fixes that.

richm:
> > I have retested using ternary `ha_cluster_fence_agent_packages: "{{ ha_cluster_fence_agent_packages | length == 0 | ternary(__ha_cluster_fence_agent_packages, ha_cluster_fence_agent_packages) }}"` and it ends with endless nesting like if/else.
> > It might be simplest to do set_facts before firewall/selinux block:
> > ```
> > - name: Set ha_cluster_fence_agent_packages fact
> >   ansible.builtin.set_fact:
> >     ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages }}"
> >   when: ha_cluster_fence_agent_packages | length == 0
> > ```
> 
> This would work, but it modifies user-defined `ha_cluster_fence_agent_packages` variable, and the modification will be in effect even after the role finishes. Users won't expect this to happen and it may cause issues. I would rather set a "private" variable. If there is endless nesting, perhaps we need a new variable name? `__ha_cluster_fence_agent_packages_final: "{{ ha_cluster_fence_agent_packages | length == 0 | ternary(__ha_cluster_fence_agent_packages_default, ha_cluster_fence_agent_packages) }}"`
> 
> Also, I would put this task into `check-and-prepare-role-variables.yml`

I think my proposal https://github.com/linux-system-roles/ha_cluster/pull/186#discussion_r1484452900 will solve this problem.

defaults/main.yml
marcelmamula: @richm I have changed to in latest commit after conversation with @tomjelinek.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
```diff
@@ -0,0 +1,135 @@
...
+      | selectattr('name', 'match', '^auto_tie_breaker$')
+      | map(attribute='value') | select('in', ['0', 0]) | list | length > 0
+
+# Sets fence agent list from os_familu vars depending on user input.
+- name: Set __ha_cluster_fence_agent_packages_final fact
+  ansible.builtin.set_fact:
+    __ha_cluster_fence_agent_packages_final:
+      "{{ (ha_cluster_fence_agent_packages | length == 0) |
+        ternary(__ha_cluster_fence_agent_packages_default,
+        ha_cluster_fence_agent_packages) }}"
```
richm, line 135: This is exactly what my proposal does, without having to resort to `set_fact` and defining yet another variable.

defaults/main.yml
richm: Looks like the default variable was renamed to `__ha_cluster_fence_agent_packages`, so this should be
```yaml
ha_cluster_fence_agent_packages: "{{ __ha_cluster_fence_agent_packages_default }}"
```

tasks/shell_crmsh/check-and-prepare-role-variables.yml
tomjelinek: Oh, yes. Somehow, I missed your comment about this. @marcelmamula, can you try implementing @richm's proposal regarding this?

marcelmamula: defaults/main.yml does not contain any __private variables. Do you want to expose them there now?
Latest commit has option for users to define fence agents and then decide based on that decision, without having them to to go and replace this assignment in defaults.

richm: > defaults/main.yml does not contain any __private variables. Do you want to expose them there now?

Yes.  This is quite a common idiom in system roles.

> Latest commit has option for users to define fence agents and then decide based on that decision, without having them to to go and replace this assignment in defaults.

Not sure what you mean.  With my proposal:
* if the user defines `ha_cluster_fence_agent_packages` then the role will use that definition
* if the user does not define `ha_cluster_fence_agent_packages` then `ha_cluster_fence_agent_packages` will be set to the default value `__ha_cluster_fence_agent_packages_default` and the default value will be used

richm: See also https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#ansible-variable-precedence

marcelmamula: @richm Both solution works, but I will update it so we can move past this.

defaults/main.yml
marcelmamula: @richm Changes are completed and pushed.

tasks/shell_crmsh/check-and-prepare-role-variables.yml
marcelmamula: [citest]
Changes are pushed for

sean-freeman (approved):
lgtm 👍

richm:
[citest] is only allowed by role maintainers or linux-system-roles org admins, for security purposes

defaults/main.yml
sean-freeman: @richm this can be marked as resolved?

tasks/shell_crmsh/check-and-prepare-role-variables.yml
sean-freeman: @richm this can be marked as resolved?

richm (approved):
thanks for being so patient!  hopefully the next PR won't require 134 gh pr comments . . .

PR #192, opened 2024-02-14
feat: SLES15 enablement, HAE detection

marcelmamula:
Enhancement: Adds additional steps for SLES15 enablement:
- High Availability Extension detection
- Package list updated to remove SLES4SAP specific package
- Cluster destroy adjusted to avoid failure when corosync-qdevice was not present
- Removed logd check, because it is part of cluster packages and it should not be checked before packages are installed on SLES15
- Added gather facts for NIC for corosync template.

Additionally main.yml was reordered to allow variable checks before package installation.

Readme was updated to note higher Service Pack requirement for both SLES and SLES4SAP, because of detected dependency on pattern patterns-ha-ha_sles, which failed to install using package module because of samba client version mismatch.

Reason: Adding coverage for SLES15 after cluster deployment verification.

Result: Testing completed on both SLES4SAP 15 SP5 and SLES15 SP5 with HA, cluster creation and destroy are working.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -8,6 +8,9 @@
   when:
     - ha_cluster_enable_repos
 
+- name: Check and prepare role variables
+  include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml  # yamllint disable-line rule:line-length
+
 - name: Install role essential packages
   package:
     name: "{{
@@ -17,9 +20,6 @@
     state: present
     use: "{{ (__ha_cluster_is_ostree | d(false)) |
              ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
-
-- name: Check and prepare role variables
-  include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml  # yamllint disable-line rule:line-length
 
 # The user is created by installing pacemaker packages. We just need to set the
 # password.
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -8,9 +8,6 @@
   when:
     - ha_cluster_enable_repos
 
-- name: Check and prepare role variables
-  include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml  # yamllint disable-line rule:line-length
-
 - name: Install role essential packages
   package:
     name: "{{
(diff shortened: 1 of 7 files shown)


tomjelinek:
[citest]

tasks/main.yml
```diff
@@ -8,6 +8,9 @@
   when:
     - ha_cluster_enable_repos
 
+- name: Check and prepare role variables
+  include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/check-and-prepare-role-variables.yml  # yamllint disable-line rule:line-length
+
```
tomjelinek, line 13: Revert this change, please. `shell_pcs/check-and-prepare-role-variables.yml` requires pcs to be installed. Hence the order: enable / check repositories, install essential packages (i.e. pacemaker shell), check and prepare variables.

tasks/shell_crmsh/cluster-destroy-crm.yml
```diff
@@ -23,6 +27,7 @@
     - pacemaker
     - corosync
     - corosync-qdevice
+  when: item in ansible_facts.services
```
tomjelinek, line 30: Are you sure that this works? On my systems, service names in `ansible_facts.services` have `.service` suffix.

tomjelinek (changes requested):

tasks/shell_crmsh/cluster-destroy-crm.yml
marcelmamula: @tomjelinek Ansible service_facts documentation shows that variable is ansible_facts.services
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/service_facts_module.html

tasks/main.yml
marcelmamula: @tomjelinek all ha_cluster packages (including qdevice setup) are coming from HA Extension so we cannot even run Install Essentials before check-and-prepare-role-variables.yml.

One option would be to empty out __ha_cluster_role_essential_packages and move all of them to __ha_cluster_fullstack_node_packages instead?

Or sneak this HAE detection into previous step enable-package-repositories.yml snce it is not used for Suse.

tomjelinek: > Or sneak this HAE detection into previous step `enable-package-repositories.yml` since it is not used for Suse.

Yes, please. That's where this belongs, that's the purpose of that step.

marcelmamula: @tomjelinek Thank you, i have reverted that ordering and moved it to repo steps.

tasks/shell_crmsh/cluster-destroy-crm.yml
spetrosi: @tomjelinek means that service names should be suffixed with .service, like `pacemaker.service`, `corosync.service`, `corosync-qdevice.service`.

marcelmamula: @spetrosi @tomjelinek Thank you for clarification, I have updated service names with .service and retested.

tomjelinek (approved):

tomjelinek:
[citest]

richm (approved):

PR #197, opened 2024-03-27
feat: crmsh enhancements, master slave, validations

marcelmamula:
Enhancement:
- Master Slave clone resource added for crmsh to support retroactively older versions. Readme updated with new inputs.
- New validations added for CIB steps to ensure that constraints are created on top of resources that exist, to avoid crmsh getting stuck.
- Expect for maintenance mode switch was added to handle resource attributes.
- New naming convention for resources, because crmsh does not generate default without providing id.
- cluster_destroy: file based loop replaced with shell module for faster cleanup.

Reason:
- Support for current=older installations using SAPHanaSR resource agent with Master Slave clone resource
- Validations to avoid crmsh getting stuck

Result:
- Cluster build was successfully tested on SLES4SAP15 SP3/5/6 and RHEL8.6
- All non-readme changes were done under crmsh so pcs is unaffected.


tasks/shell_crmsh/crm-cib-constraint-colocation.yml
```diff
@@ -1,11 +1,33 @@
 # SPDX-License-Identifier: MIT
 ---
+# Verify if constraint.resource_leader.id exists
+- name: Verify resource_leader presence {{ constraint.resource_leader.id }}
+  ansible.builtin.command:
+    cmd: |
```
richm, line 6:
```suggestion
    cmd: >-
```
?  That is, should the command be one line `crm -c {{ __ha_cluster_crm_shadow }} configure show {{ constraint.resource_leader.id }}` ?  because the yaml string `|` operator will preserve newlines, so the command will have newlines in it

tasks/shell_crmsh/crm-cib-constraint-colocation.yml
```diff
@@ -1,11 +1,33 @@
...
+      configure show {{ constraint.resource_leader.id }}
+  register: __ha_cluster_constraint_resource_leader
+  changed_when: false
+  failed_when:
+    - "'does not exist' in __ha_cluster_constraint_resource_leader.stderr"
+
+# Verify if constraint.resourceresource_follower_leader.id exists
+- name: Verify resource_follower presence {{ constraint.resource_follower.id }}
+  ansible.builtin.command:
+    cmd: |
```
richm, line 17: see above

tasks/shell_crmsh/crm-cib-constraint-colocation.yml
```diff
@@ -21,7 +43,7 @@
 - name: Delete present colocation constraint {{ __ha_cluster_constraint_id }}
   ansible.builtin.command:
     cmd: |
```
richm, line 45: see above

tasks/shell_crmsh/crm-cib-constraint-colocation.yml
marcelmamula: @richm I have run it with pipes too many times and it never resulted in error, crmsh was able to handle multi line input.

tasks/shell_crmsh/crm-cib-constraint-location.yml
```diff
@@ -1,13 +1,26 @@
 # SPDX-License-Identifier: MIT
 ---
+# Verify if constraint.resource.id exists
+- name: Verify resource presence {{ constraint.resource.id }}
+  ansible.builtin.command:
+    cmd: |
```
richm, line 6: see above

tasks/shell_crmsh/crm-cib-constraint-location.yml
```diff
@@ -23,7 +36,7 @@
 - name: Delete present location constraint {{ __ha_cluster_constraint_id }}
   ansible.builtin.command:
     cmd: |
```
richm, line 38: see above

tomjelinek (changes requested):
Master / slave resources have been deprecated in Pacemaker 2.0 released in July 2018 and replaced by promotable clone resources. In Pacemaker 3, which is expected to be released this year, they are going to be removed completely. I see no reason for adding a special variable for defining those.

The role already supports configuring promotable clones by the means of `promotable` option in `ha_cluster_resource_clones`. What would be a difference between a master_slave_clone and a promotable clone? I don't see any difference in the variable structure, except the `promotable` key. It looks to me like you are creating a new variable for something that can be defined using already existing variables.

If for whatever reason you need to create master / slave resources in pacemaker, I suggest using clones with `promotable: true` to define those. It would be then up to the crmsh portion of the role or the crmsh itself to decide if a promotable clone or a master / slave should be created based on pacemaker or crmsh version. That way, it will be transparent for the role users and the same playbook could be used for both old and new versions of pacemaker as well as both pcs-oriented and crmsh-oriented distros. Such interoperability is one of the goals of the system roles.

marcelmamula:
> Master / slave resources have been deprecated in Pacemaker 2.0 released in July 2018 and replaced by promotable clone resources. In Pacemaker 3, which is expected to be released this year, they are going to be removed completely. I see no reason for adding a special variable for defining those.
> 
> The role already supports configuring promotable clones by the means of `promotable` option in `ha_cluster_resource_clones`. What would be a difference between a master_slave_clone and a promotable clone? I don't see any difference in the variable structure, except the `promotable` key. It looks to me like you are creating a new variable for something that can be defined using already existing variables.
> 
> If for whatever reason you need to create master / slave resources in pacemaker, I suggest using clones with `promotable: true` to define those. It would be then up to the crmsh portion of the role or the crmsh itself to decide if a promotable clone or a master / slave should be created based on pacemaker or crmsh version. That way, it will be transparent for the role users and the same playbook could be used for both old and new versions of pacemaker as well as both pcs-oriented and crmsh-oriented distros. Such interoperability is one of the goals of the system roles.

Master Slave resources were indeed deprecated, but SUSE still uses Master Slave for existing clusters. It is replaced by Clone promotable in new Resource Agent SAPHanaSR-angi, but all current SAPHanaSR implementations are documented and recommended to use Master Slave.

This is added to ensure proper support coverage and it has no impact on pcs.

marcelmamula:
> > @tomjelinek Master Slave resources were indeed deprecated, but SUSE still uses Master Slave for existing clusters. It is replaced by Clone promotable in new Resource Agent SAPHanaSR-angi, but all current SAPHanaSR implementations are documented and recommended to use Master Slave.
> > This is added to ensure proper support coverage and it has no impact on pcs.
> 
> @marcelmamula I'm not arguing that master / slave resources do not need to be supported. My point is different: What is preventing you to use already existing `promotable: true` syntax for defining master / slave resources? Something like this:
> 
> ```
> crm configure {{ resource.promotable | ternary("ms", "clone") }} {{ resource.id }} ...
> ```
> 
> I don't see any other difference between promotable clones and master / slave resources than this.

@tomjelinek My idea was to have new code for different type of resource and have it documented.
But your proposal makes sense. I will try to redo it tomorrow and see how it behaves.

marcelmamula:
@tomjelinek I can simplify this msl scenario with hidden attribute under clone dictionary, that would determine ms/clone conditional. There is currently no difference for 2 of our clones in terms of promotable, as it is undefined so it cannot be used for decision. I would love to avoid using id as decision factor.

it can be hidden atribute like ms, master, msl as boolean.

```
ha_cluster_resource_master_slave_clones:
  - id: msl_SAPHana_H00_HDB00
    resource_id: rsc_SAPHana_H00_HDB00
    meta_attrs:
      - attrs:
          - name: clone-max
            value: 2
          - name: clone-node-max
            value: 1
          - name: interleave
            value: true
          - name: maintenance
            value: true
    ms: true
```

tasks/shell_crmsh/crm-cib-constraint-colocation.yml
marcelmamula: @richm  I have changed all pipes where it was necessary, change is in latest commit.

marcelmamula:
@tomjelinek master slave code is removed and existing clone is adjusted to expect clone attribute ms: true for this specific use case. It would stay with default clone if it is not used.

tasks/shell_crmsh/crm-cib-constraint-ticket.yml
```diff
@@ -31,9 +43,18 @@
     cmd: |
```
richm, line 43:
```suggestion
    cmd: >-
```

marcelmamula: @richm I am keeping pipe in configure commands as they can be multi-line like primitives and it works. I have removed it for one line commands above them.

tasks/shell_crmsh/crm-cib-resource-clone.yml
```diff
@@ -20,24 +31,31 @@
...
   when: __ha_cluster_resource_status.rc == 0
   check_mode: false
   changed_when: not ansible_check_mode
 
+# Clone is default resource type, unless attribute ms: true is specified
+# Pacemaker 2.0 deprecated use of ms, but it is still valid and supported
+# solution for SAP Hana clusters on SUSE using SAPHanaSR.
 - name: Configure resource clone {{ __ha_cluster_resource_id }}
   ansible.builtin.command:
     cmd: |
```
richm, line 46:
```suggestion
    cmd: >-
```

tasks/shell_crmsh/crm-cib-resource-primitive.yml
```diff
@@ -47,5 +49,7 @@
           {% endif %}
         {% endfor %}
       {% endif %}
+  # crm can get stuck if it encounters error and expects prompt response.
+  timeout: 60
```
richm, line 53: What is this?   Where is the behavior of the `timeout` keyword documented?

marcelmamula: @richm Timeout is ansible task parameter. I have added it because crmsh gets stuck if it encounters error and asks for user input. This is handled with expect in constraint tasks, while keeping resource tasks with timeout.

tasks/shell_crmsh/crm-cib-constraint-ticket.yml
marcelmamula: @richm I kept it because lot of configure tasks contain lot of operations and additional parameters so they are loaded with newline character \n. This is reason why I kept it with pipe. 

Situation is not requiring multi-line input for checks and deletions of course, only configure itself has lot of inputs.

All this code was retested many times and crmsh is able to take multi-line input correctly.

tasks/shell_crmsh/crm-cib-resource-primitive.yml
richm: ok, I see - https://docs.ansible.com/ansible/latest/reference_appendices/config.html#task-timeout
"Version Added:   2.10"

Note that for RedHat os_family, we still have to support Ansible 2.9, and still have many customers using 2.9, until late 2024 when 2.9 support will be ended.  Not sure if you have the same restrictions.

marcelmamula: @richm I was doing testing on my execution node with 2.16 so I was not checking version of it.
It seemed to fulfill task of making sure it does not hang forever. Only other way I see to handle it would be using async and then async_status instead of it.

Do you know of other way to ensure that command/shell does not get stuck like this? I dont see any option with crmsh to skip or auto-decline user inputs without using command/shell with echo pipe.

richm: > @richm I was doing testing on my execution node with 2.16 so I was not checking version of it. It seemed to fulfill task of making sure it does not hang forever. Only other way I see to handle it would be using async and then async_status instead of it.
> 
> Do you know of other way to ensure that command/shell does not get stuck like this? I dont see any option with crmsh to skip or auto-decline user inputs without using command/shell with echo pipe.

I would say - use `timeout:` unless you are sure that you will have some customers using Ansible 2.9
If you must support Ansible 2.9, then you will need to use async

tasks/shell_crmsh/crm-cib-constraint-colocation.yml
marcelmamula: Fixed

tasks/shell_crmsh/crm-cib-constraint-location.yml
marcelmamula: Fixed

tasks/shell_crmsh/crm-cib-constraint-location.yml
marcelmamula: Fixed

tasks/shell_crmsh/crm-cib-resource-clone.yml
marcelmamula: @richm I have changed pipe to >- where it was possible and I am keeping it for multi-line commands only. Can we close pipe discussion?

tasks/shell_crmsh/crm-cib-resource-primitive.yml
marcelmamula: @richm I have removed timeout and replaced it with expect module with response y/n and ERROR.

marcelmamula:
@richm @tomjelinek I will be pushing one more commit with new functionality that was on todo list for crmsh: rsc/op defaults.

tomjelinek:
[citest]

tomjelinek (approved):
Thank you for updating the PR, it looks good to me now. Let's wait for the CI.

richm (approved):

PR #201, opened 2024-04-17
fix: make consistent approach for multiple node attributes sets

idevat:
Enhancement:
Handle node attributes with multiple attributes sets consistently with rest of the role.

Reason:
Node attributes multiple sets was handled in a manner inconsistent with the rest of the role. Instead of using only the first set and ignoring the rest, multiple sets was merged into one single set. Such inconsistency can confuse a user and may complicate the possible future addition of support for multiple sets.

Result:
Node attributes multiple set are handled consistently with the rest of the role. This behavior is also described more explicitly in README.md.

https://issues.redhat.com/browse/RHEL-33076

### DIFF
--- a/tasks/shell_pcs/pcs-node-attributes.yml
+++ b/tasks/shell_pcs/pcs-node-attributes.yml
@@ -7,10 +7,8 @@
     cmd: >
       pcs -f {{ __ha_cluster_tempfile_cib_xml.path | quote }}
       -- node attribute {{ node_options.node_name | quote }}
-      {% for attr_set in node_options.attributes | d([], true) %}
-        {% for attr in attr_set.attrs | d([]) %}
-          {{ attr.name | quote }}={{ attr.value | quote }}
-        {% endfor %}
+      {% for attr in node_options.attributes[0].attrs | d([]) %}
+        {{ attr.name | quote }}={{ attr.value | quote }}
       {% endfor %}
   # We always need to create CIB to see whether it's the same as what is
   # already present in the cluster. However, we don't want to report it as a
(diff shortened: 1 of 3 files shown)


tasks/shell_pcs/pcs-node-attributes.yml
```diff
@@ -7,10 +7,8 @@
     cmd: >
       pcs -f {{ __ha_cluster_tempfile_cib_xml.path | quote }}
       -- node attribute {{ node_options.node_name | quote }}
-      {% for attr_set in node_options.attributes | d([], true) %}
-        {% for attr in attr_set.attrs | d([]) %}
-          {{ attr.name | quote }}={{ attr.value | quote }}
-        {% endfor %}
+      {% for attr in node_options.attributes[0].attrs | d([]) %}
```
richm, line 10:
```suggestion
      {% for attr in node_options.attributes[0].attrs | d([]) %}
```
Not sure what will happen here if `attributes` is unset or `[]` - you might need to protect this loop with an `if`

```
      {% if node_options.attributes | d([]) | length > 0 %}
        {% for attr in node_options.attributes[0].attrs | d([]) %}
        ...
      {% endif %}
```

tomjelinek: If `attributes` is unset, the task isn't included and executed at all:
```yaml
    - name: Configure node attributes
      include_tasks: pcs-node-attributes.yml
      loop: "{{ ha_cluster_node_options | selectattr('attributes', 'defined') |
        list }}"
      loop_control:
        loop_var: node_options
```
If `attributes` is an empty list, then `d([])` catches that and returns an empty list. I just tested this with ansible 2.15.

There might be an issue in case `attrs` is an empty list or `d` returns an empty list. In that case, the pcs command would be `pcs -f {tmpfile} -- node attribute {node-name}`. Currently, it is not an issue, since the command doesn't fail, and that is unlikely to change in future. But to be on the safe side, and also because the command doesn't change cluster configuration, `when: node_options.attributes[0].attrs | d([]) | length > 0` could be added to the task.

tomjelinek:
[citest]

tomjelinek (approved):

PR #208, opened 2024-06-05
ci: fix Python discovery with Ansible 2.17+

tomjelinek:
Fix Python discovery with Ansible 2.17+ in gitlab CI. Also, make the gitlab CI script more verbose and replace if-then-else with && or ||.

### DIFF
--- a/.gitlab-ci-tier1.yml.j2
+++ b/.gitlab-ci-tier1.yml.j2
@@ -10,19 +10,25 @@
 # https://github.com/ansible/awx/issues/14496#issuecomment-1743711473
 .run_test:
   before_script:
-    - source "${PYENV_ROOT}/bin/activate"
-    - python --version
-  script:
     - echo "$BASE_IMAGE_NAME"
     - echo "$TEST_NAME"
-    - if [ "x${BASE_IMAGE_NAME}" == "xLsrRhel8OldestAnsibleCurrent" ]; then GALAXY_SERVER="https://old-galaxy.ansible.com"; fi
+    - source "${PYENV_ROOT}/bin/activate"
+    - python --version
+    - ansible --version
+  script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci-tier1.yml.j2
```diff
@@ -10,19 +10,25 @@
...
-    - echo "$BASE_IMAGE_NAME"
-    - echo "$TEST_NAME"
-    - if [ "x${BASE_IMAGE_NAME}" == "xLsrRhel8OldestAnsibleCurrent" ]; then GALAXY_SERVER="https://old-galaxy.ansible.com"; fi
+    - ([ "x${BASE_IMAGE_NAME}" == "xLsrRhel8OldestAnsibleCurrent" ]) && GALAXY_SERVER="https://old-galaxy.ansible.com" || GALAXY_SERVER=""
     - echo "$GALAXY_SERVER"
     - ansible-galaxy collection install -vvv "--server=${GALAXY_SERVER}" -r ./meta/collection-requirements.yml
     - varsparams=""
+    # Use ansible-2.16-like Python in ansible 2.17+ (RHEL8 supports ansible < 2.17 only)
+    # See https://github.com/linux-system-roles/ha_cluster/pull/205#issuecomment-2141431543
+    - (echo "x${BASE_IMAGE_NAME}" | grep -Eq '^xLsrRhel9.+AnsibleCurrent$') && varsparams+=" -e ansible_python_interpreter=/usr/bin/python3"
```
richm, line 25: In bash, I don't think you can set a variable in a sub-shell and have that variable propagate to the parent - same with setting a variable on the right hand side of a `|` pipe

tomjelinek: I'm not sure what you mean. The `(echo ... | grep ...)` part is there only to exit with 0 or 1, which causes `varsparams+=` to either happen or not happen. If that's what you mean.

I already tested this code, it sets `varsparams` correctly - I put `echo` bellow to check that.

richm (approved):

PR #223, opened 2024-08-19
fix: Fixes for new pcs and ansible

tomjelinek:
This makes the role work with the upcoming version of pcs and cluster stack in general.

It also addresses the removal of password_hash filter due to removal of passlib.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -23,14 +23,26 @@
 
 # The user is created by installing pacemaker packages. We just need to set the
 # password.
-- name: Set hacluster password
-  user:
-    name: hacluster
-    password: "{{
-      ha_cluster_hacluster_password | string |
-      password_hash('sha512', ansible_hostname.replace('-', 'x')[:16]) }}"
+- name: Provide a password for the hacluster user
   when:
     - ha_cluster_hacluster_password | string | length > 0
+  block:
+    - name: Generate a password hash
+      # The arg `-6` means SHA512 based algorithms.
+      command:
+        cmd: >-
+          openssl passwd
+            -6
+            -salt {{ ansible_hostname.replace('-', 'x') | quote }}
+            {{ ha_cluster_hacluster_password | string | quote }}
+      register: __ha_cluster_openssl_call_result
+      changed_when: false
+      no_log: true
+
+    - name: Set hacluster password
+      user:
+        name: hacluster
+        password: "{{ __ha_cluster_openssl_call_result.stdout }}"
 
 - name: Configure shell
   include_tasks: shell_{{ ha_cluster_pacemaker_shell }}/configure-shell.yml  # yamllint disable-line rule:line-length
--- a/tasks/shell_pcs/pcs-cib-resource-primitive.yml
+++ b/tasks/shell_pcs/pcs-cib-resource-primitive.yml
@@ -65,14 +65,14 @@
     cmd: >
       pcs -f {{ __ha_cluster_tempfile_cib_xml.path | quote }}
       -- resource utilization {{ resource.id | quote }}
-      {% if resource.utilization | d([]) %}
-        {% for attr in resource.utilization[0].attrs | d([]) %}
-          {{ attr.name | quote }}={{ attr.value | quote }}
-        {% endfor %}
-      {% endif %}
+      {% for attr in resource.utilization[0].attrs | d([]) %}
+        {{ attr.name | quote }}={{ attr.value | quote }}
+      {% endfor %}
   # We always need to create CIB to see whether it's the same as what is
   # already present in the cluster. However, we don't want to report it as a
   # change since the only thing which matters is pushing the resulting CIB to
   # the cluster.
   check_mode: false
   changed_when: not ansible_check_mode
+  when:
+    - resource.utilization | d([])
--- a/tests/tests_cib_constraints_create.yml
+++ b/tests/tests_cib_constraints_create.yml
@@ -563,7 +563,7 @@
           when:
             - '"pcmk.constraint.config.output-formats" in __test_pcs_capabilities'
           vars:
-            __test_expected_lines: '
+            __test_expected_lines: |
                 {
                     "colocation": [],
                     "colocation_set": [],
@@ -1538,7 +1538,7 @@
                     "order_set": [],
                     "ticket": [],
                     "ticket_set": []
-                }'
+                }
           block:
             - name: Fetch location constraints configuration from the cluster
               command:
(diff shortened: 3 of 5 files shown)


tomjelinek:
[citest]

tasks/main.yml
```diff
@@ -23,14 +23,27 @@
...
+      # The arg `-6` means SHA512 based algorithms.
+      command:
+        cmd: >-
+          openssl passwd
+            -6
+            -salt {{ ansible_hostname.replace('-', 'x') }}
+            {{ ha_cluster_hacluster_password | string }}
+      register: __ha_cluster_openssl_call_result
+      changed_when: false
+      failed_when: __ha_cluster_openssl_call_result.rc != 0
```
spetrosi, line 40: I think this is the default Ansible behavior, so need to state this `failed_when`.

tasks/main.yml
```diff
@@ -23,14 +23,27 @@
...
-      ha_cluster_hacluster_password | string |
-      password_hash('sha512', ansible_hostname.replace('-', 'x')[:16]) }}"
+- name: Provide a password for the hacluster user
   when:
     - ha_cluster_hacluster_password | string | length > 0
+  block:
+    - name: Generate a password hash
+      # The arg `-6` means SHA512 based algorithms.
+      command:
+        cmd: >-
```
spetrosi, line 33: No need to use `cmd` parameter here

```
      command: >-
        openssl passwd
          -6
          -salt {{ ansible_hostname.replace('-', 'x') }}
          {{ ha_cluster_hacluster_password | string }}
```

tasks/shell_pcs/pcs-cib-resource-primitive.yml
```diff
@@ -65,14 +65,14 @@
     cmd: >
       pcs -f {{ __ha_cluster_tempfile_cib_xml.path | quote }}
       -- resource utilization {{ resource.id | quote }}
-      {% if resource.utilization | d([]) %}
```
spetrosi, line 68: Why do you remove this `if`?

spetrosi (commented):
Thanks for removing dependency for `password_hash`

tasks/shell_pcs/pcs-cib-resource-primitive.yml
tomjelinek: I moved it to `when` statement. If there is no utilization defined, we don't want the command to be run at all.

tasks/main.yml
tomjelinek: Thanks, I removed the line.

tomjelinek:
[citest]

tasks/main.yml
spetrosi: Ansible supports providing the cmd parameter right with `command`, if `cmd` is the only parameter that you need to use.

tomjelinek:
Centos-8 fails with an error that looks like a CI issue to me:
```
TASK [List packages on the controller to see if OpenSSL is installed] **********
task path: /tmp/collections-1sG/ansible_collections/fedora/linux_system_roles/tests/ha_cluster/tasks/fixture_psks.yml:15
Monday 19 August 2024  06:43:32 -0400 (0:00:00.078)       0:00:02.210 ********* 
[WARNING]: Found "rpm" but Failed to import the required Python library (rpm)
on ip-10-31-46-26.us-east-1.aws.redhat.com's Python /usr/bin/python3.9. Please
read module documentation and install in the appropriate location. If the
required library is installed, but Ansible is using the wrong Python
interpreter, please consult the documentation on ansible_python_interpreter
fatal: [managed_node2 -> localhost]: FAILED! => {
    "changed": false
}

MSG:

Could not detect a supported package manager from the following list: ['rpm', 'apt', 'portage', 'pkg'], or the required Python library is not installed. Check warnings for details.
```

tasks/main.yml
```diff
@@ -23,14 +23,26 @@
...
+          openssl passwd
+            -6
+            -salt {{ ansible_hostname.replace('-', 'x') }}
+            {{ ha_cluster_hacluster_password | string }}
```
richm, line 37:
```suggestion
            {{ ha_cluster_hacluster_password | string | quote }}
```

tasks/main.yml
```diff
@@ -23,14 +23,26 @@
...
   when:
     - ha_cluster_hacluster_password | string | length > 0
+  block:
+    - name: Generate a password hash
+      # The arg `-6` means SHA512 based algorithms.
+      command:
+        cmd: >-
+          openssl passwd
+            -6
+            -salt {{ ansible_hostname.replace('-', 'x') }}
```
richm, line 36:
```suggestion
            -salt {{ ansible_hostname.replace('-', 'x') | quote }}
```

tasks/main.yml
```diff
@@ -23,14 +23,26 @@
...
+            {{ ha_cluster_hacluster_password | string }}
+      register: __ha_cluster_openssl_call_result
+      changed_when: false
+      no_log: true
+
+    - name: Set hacluster password
+      user:
+        name: hacluster
+        password: "{{ __ha_cluster_openssl_call_result.stdout }}"
```
richm, line 45:
```suggestion
        password: "{{ __ha_cluster_openssl_call_result.stdout }}"
      no_log: true
```

tests/tests_cib_constraints_create.yml
```diff
@@ -563,7 +563,7 @@
           when:
             - '"pcmk.constraint.config.output-formats" in __test_pcs_capabilities'
           vars:
-            __test_expected_lines: >
+            __test_expected_lines: '
```
richm, line 566: I guess this will work - but the usual way to do a multiline string in yaml is to use a literal block scalar with `|`

tasks/main.yml
tomjelinek: I can add no_log, but is it necessary? The password is not shown, the user module hides it even without no_log:
```
TASK [linux-system-roles.ha_cluster : Set hacluster password] ******************
changed: [localhost] => {
    "append": false,
    "changed": true,
    "comment": "cluster user",
    "group": 189,
    "home": "/var/lib/pacemaker",
    "move_home": false,
    "name": "hacluster",
    "password": "NOT_LOGGING_PASSWORD",
    "shell": "/sbin/nologin",
    "state": "present",
    "uid": 189
}
```

tasks/main.yml
tomjelinek: Good catch, thanks! Fixed.

tasks/main.yml
tomjelinek: Good catch, thanks! Fixed.

tests/tests_cib_constraints_create.yml
tomjelinek: This needed few fixes in the following tasks, but I made it work. Thanks.

tomjelinek:
[citest]

richm:
The `CentOS-Stream-8|ansible-2.9` failures are not related to this PR.

PR #227, opened 2024-08-21
fix: openssl is now required for password

richm:
openssl is now used to generate passwords, so make it
a required basic package.


richm:
@marcelmamula this needs to change on Suse also - not sure if I have the right package name

tomjelinek (approved):

richm:
[citest]

marcelmamula:
> @marcelmamula this needs to change on Suse also - not sure if I have the right package name

@richm Package name is correct. LGTM

PR #231, opened 2024-10-08
feat: export corosync configuration

tomjelinek:
Enhancement:
Provide `ha_cluster_info` module to export current cluster configuration. This PR implements a first stage, exporting corosync configuration. Other parts of configuration will follow in other PRs.

Reason:
This is the first step in implementing an info module which exports cluster configuration in a variables structure in the same format as ha_cluster role accepts.

Result:
ha_cluster_info module exports corosync configuration, which can be used to recreate the same corosync cluster when passed to the role

https://issues.redhat.com/browse/RHEL-46219


tomjelinek:
[citest]

tomjelinek:
I updated the pcs_version vs ubuntu version matrix, as pcs main no longer builds on ubuntu-22.04. And then Python Unit Tests / python (ubuntu-24.04, main) fails when trying to upgrade pip:
```
+ python -m pip install --upgrade pip
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```
I suppose upgrading pip could be removed. But I'm afraid that would solve nothing, as the next command is `pip install "git+https://github.com/linux-system-roles/tox-lsr@3.4.0"` and that would probably fail with the same message. So what's the process of installing tox-lsr on ubuntu-24.04?

tomjelinek:
ansible_test fails with this error:
```
Running sanity test "ansible-doc"
Run command: ansible-doc -t module fedora.linux_system_roles.ha_cluster_info fedora.linux_system_roles.pcs_api_v2 fedora.linux_system_roles.pcs_qdevice_certs
ERROR: Output on stderr from ansible-doc is considered an error.

Command "ansible-doc -t module fedora.linux_system_roles.ha_cluster_info fedora.linux_system_roles.pcs_api_v2 fedora.linux_system_roles.pcs_qdevice_certs" returned exit status 0.
>>> Standard Error
Warning: : Collection fedora.linux_system_roles does not support Ansible
version 2.14.17.post0
```
Any idea what this means and how to fix it?

tomjelinek:
CentOS-Stream-8|ansible-2.9 fails with `Could not detect a supported package manager from the following list: ['pkg', 'apt', 'rpm', 'portage'], or the required Python library is not installed. Check warnings for details.`. I think we went over this already, and the resolution was that this was an incompatibility between CentOS 8 and Ansible.

I'm not sure why the other CentOS and Fedora tests are marked as failures, when all their logs are success.

spetrosi:
[citest]

spetrosi:
> CentOS-Stream-8|ansible-2.9 fails with `Could not detect a supported package manager from the following list: ['pkg', 'apt', 'rpm', 'portage'], or the required Python library is not installed. Check warnings for details.`. I think we went over this already, and the resolution was that this was an incompatibility between CentOS 8 and Ansible.

Looking into this, I think it used to work, idk what broke it.

> I'm not sure why the other CentOS and Fedora tests are marked as failures, when all their logs are success.

Fixed in https://github.com/linux-system-roles/tft-tests/pull/53, tests passed but some tasks run in background after the testing phase finished, it caused the failure of test plan. Now it's passing.

spetrosi:
Fixing issue with ansible-2.9 on CS8 in https://github.com/linux-system-roles/tft-tests/pull/54

spetrosi:
[citest]

README.md
```diff
@@ -1545,6 +1551,79 @@ all:
...
+    responsibility to decide if you want to use existing keys or generate new
+    ones.
+
+To export current cluster configuration and store it in
+`ha_cluster_info_result` variable, write a task like this:
+
+```yaml
+- name: Get current cluster configuration
+  linux-system-roles.ha_cluster.ha_cluster_info:
+  register: ha_cluster_info_result
```
richm, line 1599: System roles, by convention, do not support users using modules directly.  In every other role that does something like this, users use the role with either no arguments like https://github.com/linux-system-roles/firewall?tab=readme-ov-file#gathering-firewall-ansible-facts:

```yaml
- name: Get current cluster configuration
  include_role:
    name: linux-system-roles.ha_cluster
```

or with some special variable

```yaml
- name: Get current cluster configuration
  include_role:
    name: linux-system-roles.ha_cluster
  vars:
    ha_cluster_get_info: true
```

I think the ha_cluster role will have to do something like the latter, since there are numerous public api variables, as opposed to the firewall role which just has the one main `firewall` variable.  The latter also makes it possible for the role to

* set the state of the cluster *and* return the cluster configuration
* return a subset of the information

richm: The role would then set a *global* variable e.g. `ha_cluster_info` that users would use.  This return variable will be declared in the README.md in the section `Variables Exported by the Role` e.g. https://github.com/linux-system-roles/kernel_settings?tab=readme-ov-file#variables-exported-by-the-role

richm: ping

spetrosi: bootloader and snapshot roles export info with `<rolename>_facts` variable. Let's be consistent with this naming.

tomjelinek: Sorry, I've been busy with other projects.

I wasn't sure what would be the correct way to expose the export functionality. So I'm glad you pointed me in the right direction. I'm going to implement this change, hopefully in a couple of weeks, once I finish tasks that require my immediate attention.

tomjelinek:
[citest]

README.md
tomjelinek: I pushed new commits which implement this requested change.

richm (approved):
lgtm - @spetrosi ptal

README.md
```diff
@@ -59,6 +64,13 @@ ansible-galaxy collection install -r meta/collection-requirements.yml
 
 ### Defined in `defaults/main.yml`
 
+#### `ha_cluster_get_info`
```
spetrosi, line 67: I thought initially that  it makes sense to rename this to `ha_cluster_facts` for consistency with other roles. Although this functionality is a little more than just getting the current state, it exports the configuration in a format that can be used to run the role.
Maybe rename this to `ha_cluster_export_configuration`, I think it's more in the face naming, if we export the configuration it implies that we can import it with the same role, which is the case here.
@richm WDYT?

richm: Ok - sounds good - please rename to `ha_cluster_export_configuration` and then I think we're good to go.

tomjelinek:
[citest]

tomjelinek:
CentOS-10 tests_cluster_basic_cloud_packages will be fixed in another pull request. CentOS-9 tests_qdevice_tls_kaptb_options is a bit flaky test. Neither of these is related to changes done in this PR.

tomjelinek:
> Looks like the python unit tests are broken on ubuntu-24.04 - I looked at this briefly - looks like the `pip install` `--prefix` argument in that version of pip adds `/local` to the end - the Makefile then cannot find the binary to install

Yes, this needs to be fixed on pcs side. It's in the TODO list, hopefully we get to resolve it soon.

spetrosi (approved):
lgtm

PR #232, opened 2024-10-14
feat: crmsh 4.6.0 support and stonith-enabled workflow update

marcelmamula:
### main
- package installation changed from `present` to `latest` to accommodate for installation of package updates.
- Images do not come with latest maintenance patches so it is better to update to latest before building cluster.
- @richm @tomjelinek yum should support state `latest` from what I checked.
- New variable added to allow updating of packages `ha_cluster_use_latest_packages`, as discussed below. 

### crmsh
- Changed maintenance tasks to force mode (expect modules were commented out)
- Add task to disable `stonith-enabled` property before proceeding with other `crm configure` steps
- Move cluster property task to end of CIB build so stonith can be enabled at end
- Enhance cluster property task to
  - Append `stonith-enabled=true` if it is missing.
  - Show warning if it was purposefully executed with `stonith-enabled=false`, which is not recommended.
- Execute `crm_verify` to validate shadow CIB before patching. Fail module is used to check return code.

## Reason:
`crmsh 4.6.0` in SUSE Linux Enterprise Server for SAP Applications 15 SP6 no longer ignores error caused by `ha_cluster` workflow.
- crmsh should be executed by `crm init` which sets property `stonit-enabled=false`, allowing for all configuration steps including maintenance mode.
- `ha_cluster` workflow does not use `crm init` so this step was missing, resulting in RC 78 error on SP6 which is thrown for every `crm configure` command except `crm configure property stonith-enabled=false`
```console
ae1ascs:~ # crm configure property maintenance-mode=true
ERROR: error: Resource start-up disabled since no STONITH resources have been defined
ERROR: error: Either configure some or disable STONITH with the stonith-enabled option
ERROR: error: NOTE: Clusters with shared data need STONITH to ensure data integrity
ERROR: crm_verify: Errors found during check: config not valid
```

## Result:
Successfully tested building of SAP HANA and SAP Netweaver clusters on AWS and Google Cloud for OS:
- SUSE Linux Enterprise Server for SAP Applications 15 SP5
- SUSE Linux Enterprise Server for SAP Applications 15 SP6

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,8 @@ An Ansible role for managing High Availability Clustering.
 
 * Compatible OS
   * RHEL 8.3+, Fedora 31+
-  * SLES 15 SP5+ with HA extension, SLES4SAP 15 SP5+
+  * SUSE Linux Enterprise Server 15 with HA extension
+  * SUSE Linux Enterprise Server for SAP Applications 15
 * Systems running RHEL are expected to be registered and have High-Availability
   repositories accessible, and ResilientStorage repositories accessible if using
   `ha_cluster_enable_repos_resilient_storage`
@@ -152,6 +153,13 @@ automatically by the role, for example custom resource agents.
 It is possible to specify fence agents here as well. However,
 [`ha_cluster_fence_agent_packages`](#ha_cluster_fence_agent_packages) is
 preferred for that, so that its default value is overridden.
+
+#### `ha_cluster_use_latest_packages`
+
+boolean, default: `false`
+
+If set to `true`, all packages will be installed with latest version.</br>
+If set to `false`, existing packages will not be updated.
 
 #### `ha_cluster_hacluster_password`
 
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ preferred for that, so that its default value is overridden.
 
 boolean, default: `false`
 
-If set to `true`, all packages will be installed with latest version.</br>
+If set to `true`, all packages will be installed with latest version.
 If set to `false`, existing packages will not be updated.
 
 #### `ha_cluster_hacluster_password`
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,7 +14,7 @@
       __ha_cluster_role_essential_packages
       +
       ha_cluster_extra_packages }}"
-    state: present
+    state: latest
     use: "{{ (__ha_cluster_is_ostree | d(false)) |
              ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
@@ -73,7 +73,7 @@
             ternary(__ha_cluster_cloud_agents_packages, [])
           +
           ha_cluster_fence_agent_packages }}"
-        state: present
+        state: latest
         use: "{{ (__ha_cluster_is_ostree | d(false)) |
                  ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -14,7 +14,7 @@
       __ha_cluster_role_essential_packages
       +
       ha_cluster_extra_packages }}"
-    state: latest
+    state: "{{ 'latest' if ha_cluster_use_latest_packages else 'present' }}"
     use: "{{ (__ha_cluster_is_ostree | d(false)) |
              ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
@@ -73,7 +73,7 @@
             ternary(__ha_cluster_cloud_agents_packages, [])
           +
           ha_cluster_fence_agent_packages }}"
-        state: latest
+        state: "{{ 'latest' if ha_cluster_use_latest_packages else 'present' }}"
         use: "{{ (__ha_cluster_is_ostree | d(false)) |
                  ternary('ansible.posix.rhel_rpm_ostree', omit) }}"
 
--- a/tasks/shell_crmsh/create-and-push-cib.yml
+++ b/tasks/shell_crmsh/create-and-push-cib.yml
@@ -27,23 +27,28 @@
   check_mode: false
   changed_when: not ansible_check_mode
 
-# Maintenance mode is required, because CIB version changes with cluster
-# status changes, resulting in shadow CIB outdated and unable to patch.
-# Sleep is implemented to ensure that cluster have enough time to freeze
-# to ensure CIB export consistency.
 # Meta-attrs is-managed will conflict with maintenance mode as well as
 # individual resource maintenance attributes. Expect will skip their deletion.
+# - name: Put cluster in maintenance mode to freeze cib changes
+#   ansible.builtin.expect:
+#     command: crm --force configure property maintenance-mode=true
+#     responses:
+#       ".*is-managed.*": "n"
+#       ".*already.*": "n"
+#   run_once: true  # noqa: run_once[task]
+#   check_mode: false
+#   changed_when: true
+
+# Maintenance mode is required, because CIB version changes with cluster
+# status changes, resulting in shadow CIB outdated and unable to patch.
 - name: Put cluster in maintenance mode to freeze cib changes
-  ansible.builtin.expect:
-    command: crm configure property maintenance-mode=true
-    responses:
-      ".*is-managed.*": "n"
-      ".*already.*": "n"
+  ansible.builtin.command:
+    cmd: crm --force configure property maintenance-mode=true
   run_once: true  # noqa: run_once[task]
   check_mode: false
   changed_when: true
 
-- name: Verify that maintenace-mode is true
+- name: Verify that maintenance-mode is true
   ansible.builtin.command:
     cmd: crm status
   register: __ha_cluster_crm_status_maint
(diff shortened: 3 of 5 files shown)


marcelmamula:
@tomjelinek @richm Can you tell me why tests failed beside `latest` in package module?

I can move `latest` to `check-and-prepare-role-variables.yml` task file and use zypper instead, if you want to keep package module using `present`. But I would assume you would also want to have latest packages before execution.

richm:
> @tomjelinek @richm Can you tell me why tests failed beside `latest` in package module?
> 
> I can move `latest` to `check-and-prepare-role-variables.yml` task file and use zypper instead, if you want to keep package module using `present`. But I would assume you would also want to have latest packages before execution.

https://ansible.readthedocs.io/projects/lint/rules/package-latest/?h=package

tasks/main.yml
```diff
@@ -73,7 +73,7 @@
             ternary(__ha_cluster_cloud_agents_packages, [])
           +
           ha_cluster_fence_agent_packages }}"
-        state: present
+        state: latest
```
richm, line 76: If you really want to do this despite https://ansible.readthedocs.io/projects/lint/rules/package-latest/?h=package then I suggest you introduce a new variable in defaults/main.yml e.g. `ha_cluster_use_latest_packages: false` then the user can set this to `true` if they want the packages to be updated.

tasks/shell_crmsh/create-and-push-cib.yml
```diff
@@ -116,12 +121,16 @@
...
-      vars:
-        properties_set: "{{ ha_cluster_cluster_properties[0] }}"
-      when: ha_cluster_cluster_properties[0].attrs | d([])
+    ## Ensure that stonith is disabled before executing crm configure.
+    ## This is usually disabled by running crm init.
+    ## Executing crm configure without stonith results in "config not valid".
+    - name: Set property stonith-enabled to false
+      ansible.builtin.command:
+        cmd: >-
+          crm -c {{ __ha_cluster_crm_shadow }}
```
richm, line 130:
```suggestion
          crm -c {{ __ha_cluster_crm_shadow | quote }}
```
not sure if `quote` is really needed here, but if any part of `__ha_cluster_crm_shadow` comes from user input, or can contain shell metacharacters, then the use of `quote` is necessary

tasks/shell_crmsh/create-and-push-cib.yml
```diff
@@ -229,6 +238,38 @@
...
+      vars:
+        properties_set: "{{ ha_cluster_cluster_properties[0] }}"
+      when: ha_cluster_cluster_properties[0].attrs | d([])
+
+    # Verify CIB to ensure that there are no errors before applying.
+    - name: Verify shadow CIB
+      ansible.builtin.command:
+        cmd: >-
+          crm_verify -V -x
+          /var/lib/pacemaker/cib/shadow.{{ __ha_cluster_crm_shadow }}
```
richm, line 253: same here re: use of `quote`

tasks/main.yml
marcelmamula: Hmm this would mean we would have to split it into dnf/yum and zypper task files to accommodate differences in updates.

tasks/shell_crmsh/create-and-push-cib.yml
marcelmamula: this is internal variable, already used for shadow CIB commands, so quote is not needed.

tasks/shell_crmsh/create-and-push-cib.yml
marcelmamula: this is internal variable, already used for shadow CIB commands, so quote is not needed.

tasks/main.yml
marcelmamula: @richm 
I can create new task file under both pcs and crmsh, to do package install, where var would drive latest vs present.
pcs - both yum and dnf are available, but I saw yum is not updated anymore. Let me know which one would you prefer.
- yum and dnf both have state: latest and update_only: true to do this update

crmsh - zypper module has state latest as well.

tomjelinek: Upgrading cluster packages by default doesn't seem to me to be a good idea. It would probably cause cluster to be restarted on all nodes at the same time, leading to loss of service.

Why would you need to split dnf / yum / zipper? Is there anything preventing using `package` module with `state: "{{ 'latest' if ha_cluster_use_latest_packages else 'present' }}"`?

tasks/shell_crmsh/crm-cluster-properties.yml
```diff
@@ -1,15 +1,48 @@
 # SPDX-License-Identifier: MIT
 ---
+# stonith-enabled=true is required for clusters, but this task
+# allows for creation of cluster without stonith enabled.
+# It also ensures that stonith-enabled is always present.
+- name: Append stonith-enabled to ha_cluster_cluster_properties
+  ansible.builtin.set_fact:
+    __ha_cluster_property_attr: "{{
+      (properties_set.attrs + [{'name': 'stonith-enabled', 'value': 'true'}])
+      if properties_set.attrs | selectattr('name','equalto','stonith-enabled')
```
tomjelinek, line 10:
```suggestion
      if properties_set.attrs | selectattr('name', 'equalto', 'stonith-enabled')
```
To resolve a warning from ansible-lint

tomjelinek:
> @tomjelinek @richm Can you tell me why tests failed beside `latest` in package module?

Python unit tests fail because pcs-main no longer builds on ubuntu-22.04. This is being addressed in #231 

Ansible test seems to be failing due to `Warning: : Collection fedora.linux_system_roles does not support Ansible`. I'm not sure what's going on there.

tasks/shell_crmsh/crm-cluster-properties.yml
marcelmamula: @tomjelinek It would be much easier if we raised 80 char limit ... because it becomes difficult to fit all of it while making it jinja2 compatible and readable at same time.

README.md
```diff
@@ -153,6 +154,13 @@ It is possible to specify fence agents here as well. However,
 [`ha_cluster_fence_agent_packages`](#ha_cluster_fence_agent_packages) is
 preferred for that, so that its default value is overridden.
 
+#### `ha_cluster_use_latest_packages`
+
+boolean, default: `false`
+
+If set to `true`, all packages will be installed with latest version.</br>
```
spetrosi, line 161:
```suggestion
If set to `true`, all packages will be installed with latest version.

```

tasks/shell_crmsh/crm-cluster-properties.yml
```diff
@@ -7,9 +7,8 @@
   ansible.builtin.set_fact:
     __ha_cluster_property_attr: "{{
       (properties_set.attrs + [{'name': 'stonith-enabled', 'value': 'true'}])
-      if properties_set.attrs | selectattr('name','equalto','stonith-enabled')
-       | list | length == 0
-      else properties_set.attrs }}"
+      if properties_set.attrs | selectattr('name', 'equalto', 'stonith-enabled') | list | length == 0
```
spetrosi, line 10:
```suggestion
      if properties_set.attrs
      | selectattr('name', 'equalto', 'stonith-enabled') | list | length == 0
```

tasks/shell_crmsh/crm-cluster-properties.yml
```diff
@@ -7,9 +7,8 @@
...
+      else properties_set.attrs }}"  # yamllint disable-line rule:line-length
```
spetrosi, line 11:
```suggestion
      else properties_set.attrs }}"
```

tasks/shell_crmsh/crm-cluster-properties.yml
marcelmamula: This is problem with our 80 limit, making it difficult to read.

spetrosi: It depends, for those who have codespace limited to 80 symbols and need to scroll horizontally to get the full line, reading long lines is a bigger problem. I prefer complying with this requirement where possible.

tasks/shell_crmsh/crm-cluster-properties.yml
marcelmamula: completed in latest commit

tasks/shell_crmsh/crm-cluster-properties.yml
marcelmamula: completed in latest commit

README.md
marcelmamula: completed in latest commit

tasks/shell_crmsh/crm-cluster-properties.yml
marcelmamula: completed in latest commit

tasks/main.yml
marcelmamula: @tomjelinek Updated with if/else as you suggested and tested on SLES and RHEL.

richm:
> > @tomjelinek @richm Can you tell me why tests failed beside `latest` in package module?
> 
> Python unit tests fail because pcs-main no longer builds on ubuntu-22.04. This is being addressed in #231
> 
> Ansible test seems to be failing due to `Warning: : Collection fedora.linux_system_roles does not support Ansible`. I'm not sure what's going on there.

It is because of https://github.com/linux-system-roles/auto-maintenance/pull/353 - just ignore it for now

richm:
[citest]

richm (approved):

tomjelinek (approved):

spetrosi (approved):

PR #245, opened 2024-12-06
fix: update constraints commands syntax for pcs-0.12

tomjelinek:
Enhancement:
Update the role to pcs-0.12 command syntax

Reason:
New version of pcs brings updated syntax of some commands. Previous syntax is deprecated but it still works. It will be dropped eventually, though, and the role should be ready for that.

Result:
Role uses new pcs syntax if new enough pcs is available.

https://issues.redhat.com/browse/RHEL-45303


tomjelinek:
[citest]

spetrosi (approved):

PR #249, opened 2024-12-11
feat: Remove python expect dependency in crmsh

marcelmamula:
### Enhancement:

1. Remove all use of `expect` module which required Python `pexpect` module to be installed
2. Replace `expect` by using `--force` and `yes n` to skip all prompts and fail if user input is invalid or in conflict.
3. Enhance `resource_sets` handling.

### Reason:

- Extra python dependency was causing unnecessary dependencies

### Result:
- Tested on `SLES for SAP Applications 15 SP6` in AWS for SAP HANA, SAP ASCS/ERS and dummy clusters.

### DIFF
--- a/tasks/enable-repositories/Suse.yml
+++ b/tasks/enable-repositories/Suse.yml
@@ -2,26 +2,31 @@
 ---
 # All required repositories are already part of SLES for SAP 15 SP5+.
 
-# High Availability Extension is required for cluster setup as well
-# as qnetd/qdevice configuration.
-- name: Gather package facts
-  ansible.builtin.package_facts:
-    manager: auto
+# High Availability Extension (HAE) is required for cluster setup on SLES.
+# All cluster packages are present on SLES_SAP and openSUSE,
+# but not on base SLES without HAE.
+- name: Execute only on SLES
+  when: ansible_distribution.split("_")[0] == 'SLES'
+  block:
 
-- name: Check High Availability Extension presence using product file
-  ansible.builtin.stat:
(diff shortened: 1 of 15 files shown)


richm:
[citest]

richm:
please rebase on top of the latest `main` branch

marcelmamula:
> please rebase on top of the latest `main` branch

I have rebased branch and pushed it, let me know if it is alright now @richm

spetrosi:
[citest]

tomjelinek (approved):
Overall, the patch looks good to me. However, I'm not 100% sure about tasks/set_vars.yml changes. @richm, are you ok with those?

marcelmamula:
> Overall, the patch looks good to me. However, I'm not 100% sure about tasks/set_vars.yml changes. @richm, are you ok with those?

We have been using this vars loading for some time, because of how SLES_SAP is designed. This way we can separate between core SLES and SLES_SAP version without needing to create unnecessary var files for each product name.

richm:
> > Overall, the patch looks good to me. However, I'm not 100% sure about tasks/set_vars.yml changes. @richm, are you ok with those?
> 
> We have been using this vars loading for some time, because of how SLES_SAP is designed. This way we can separate between core SLES and SLES_SAP version without needing to create unnecessary var files for each product name.

+1 - vars changes lgtm - I think the role has to do this because `os_family` and/or `distribution` by themselves are not sufficient

richm (approved):

PR #254, opened 2025-01-24
fix: update for EL10

tomjelinek:


meta/main.yml
```diff
@@ -14,6 +14,7 @@ galaxy_info:
       versions:
         - "8"
         - "9"
+        - "10"
```
richm, line 17:
```suggestion
        - "10"
```
Ansible lint does not recognize "10" as a valid version.  In fact, Ansible is deprecating this version mechanism altogether.  `galaxy_tags` is now used for versions going forward.

tomjelinek: Ah, ok. Good to know. I dropped the commit changing this and kept only the one fixing the repositories.

richm (approved):

PR #265, opened 2025-03-04
fix: disable proxy when connecting to pcsd local socket

tomjelinek:
Enhancement:
Do not use proxy when connecting to local pcsd via unix socket on managed nodes.

Reason:
When a proxy is defined, it disrupts communication towards the local pcsd unix socket. This makes the role fail.

Result:
Role works even if a proxy is defined.

https://issues.redhat.com/browse/RHEL-81918

### DIFF
--- a/module_utils/ha_cluster_lsr/pcs_api_v2_utils.py
+++ b/module_utils/ha_cluster_lsr/pcs_api_v2_utils.py
@@ -214,6 +214,8 @@ def call_api_raw(
     """
     response, info = fetch_url(
         module,
+        # never use a proxy when connecting to a local unix socket (RHEL-81918)
+        use_proxy=False,
         force=True,  # do not get a cached response
         unix_socket=PCSD_SOCKET,
         url=API_ENDPOINT,


spetrosi:
[citest]

tomjelinek:
I don't get why all tests are suddenly failing with `ha_cluster_hacluster_password must be specified`. The PR is completely unrelated. Has anything changed in CI systems?

richm:
@spetrosi some problem with support for vault?  The hacluster password comes from the vault file

spetrosi:
[citest]

spetrosi:
Should be fixed in https://github.com/linux-system-roles/tft-tests/pull/70

module_utils/ha_cluster_lsr/pcs_api_v2_utils.py
```diff
@@ -214,6 +214,8 @@ def call_api_raw(
     """
     response, info = fetch_url(
         module,
+        # never use a proxy when connecting to a local unix socket (RHEL-81918)
+        use_proxy=False,
```
richm, line 218: ok - since this will only ever be a local unix socket (and not a TCP/IP socket), proxies should never be used.

richm (approved):

PR #278, opened 2025-05-12
test: update attributes and utilization tests for new pcs

tomjelinek:
Enhancement:
Update node attributes and node utilization tests so that they are able to process new output format of attributes and utilization configuration in pcs

Reason:
Tests are failing with new pcs even if the role works correctly

Result:
Tests work correctly with new pcs

https://issues.redhat.com/browse/RHEL-90078

### DIFF
--- a/tests/tests_cib_node_attributes.yml
+++ b/tests/tests_cib_node_attributes.yml
@@ -36,7 +36,12 @@
                       - name: attr2
                         value: val2
 
+        - name: Fetch versions of cluster components
+          include_tasks: tasks/fetch_versions.yml
+
         - name: Verify node attributes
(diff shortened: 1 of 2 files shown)


tomjelinek:
I'm not sure why CentOS-10 is failing to start pcsd:
`Unable to start service pcsd: A dependency job for pcsd.service failed. See 'journalctl -xe' for details.`
I suppose there is no way to see the journal?

Recently, we had an issue with pcsd not being able to start due to conflicts in curl / libcurl / pycurl packages, which prevented pcsd to start. Maybe it's the same issue?

In any case, the errors are unrelated to the changes done in this pull request.

PR #286, opened 2025-06-30
refactor: support Ansible 2.19

richm:
Ansible 2.19 introduces some big changes
https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_core_2.19.html

One big change is that data structures are no longer mutable by the use of python
methods such as `__setitem__`, `setdefault`, `update`, etc.  in Jinja constructs.
Instead, items must use filters or other Jinja operations.

One common idiom is to mutate each element in a list.  Since we cannot do this
"in-place" anymore, a common way to do this is:

```yaml
- name: Construct a new list from an existing list and mutate each element
  set_fact:
    __new_list: "{{ __new_list | d([]) + [mutated_item] }}"
  loop: "{{ old_list }}"
  mutated_item: "{{ some value based on item from old list }}"

- name: Reset original old list
  set_fact:
    old_list: "{{ __new_list }}"
```

Similarly with `dict` items:

```yaml
- name: Construct a new dict from an existing dict and mutate each element
  set_fact:
    __new_dict: "{{ __new_dict | d({}) | combine(mutated_item) }}"
  loop: "{{ old_dict | dict2items }}"
  mutated_item: "{{ {item.key: mutation of item.value} }}"

- name: Reset original old dict
  set_fact:
    old_dict: "{{ __new_dict }}"
```

Another big change is that a boolean expression in a `when` or similar construct
must be converted to a boolean - we cannot rely on the implicit evaluation in
a boolean context.  For example, if `var` is some iterable, like a `dict`, `list`,
or `string`, you used to be able to evaluate an empty value in a boolean context:

```yaml
when: var  # do this only if var is not empty
```

You now have to explicitly test for empty using `length`:

```yaml
when: var | length > 0  # do this only if var is not empty
```

For tests_cib_constraints_create.yml - for some reason the formatting
was off, and doing the comparison of the expected data structure with
the actual one was not working.  Instead, convert both to internal
data structures with `from_json` and compare the data structures.

These are the biggest changes.  See the porting guide for others.

## Summary by Sourcery

Refactor role tasks and tests for compatibility with Ansible 2.19’s strict immutability and boolean evaluation rules by converting implicit mutations and condition checks into filter-based operations and updating tests to normalize and compare JSON structures.

Bug Fixes:
- Fix formatting in location constraints tests to properly compare expected vs actual JSON via `from_json`.
- Correct header test to use the renamed `__ansible_managed` lookup variable for assertion.

Enhancements:
- Refactor all when conditions to enforce explicit boolean evaluation using length checks and `is not none` filters.
- Migrate Jinja2 data mutations to filter-based constructs (combine, map, default) in place of in-place methods for list and dict transformations.
- Simplify certificate_requests generation by mapping each dict with `combine` instead of using `__setitem__` in loops.
- Standardize JSON output in debug tasks using `to_nice_json` and sorted lists for consistent readability.

Tests:
- Convert both actual and expected outputs to parsed JSON structures with `from_json` before assertion in constraint tests.

### DIFF
--- a/tasks/shell_pcs/check-and-prepare-role-variables.yml
+++ b/tasks/shell_pcs/check-and-prepare-role-variables.yml
@@ -39,7 +39,7 @@
           (
             __nodes_from_options != (__nodes_from_options | unique)
           ) or (
-            __nodes_from_options | difference(__ha_cluster_all_node_names)
+            __nodes_from_options | difference(__ha_cluster_all_node_names) | list | length > 0
           )
 
 - name: Extract node options
@@ -166,7 +166,7 @@
   when:
     - ha_cluster_pcsd_public_key_src is not none
     - ha_cluster_pcsd_private_key_src is not none
-    - ha_cluster_pcsd_certificates | d([])
+    - ha_cluster_pcsd_certificates | d([]) | length > 0
 
 - name: Fetch pcs capabilities
   command:
@@ -216,11 +216,11 @@
       defaults. Please, upgrade pcs.
   when:
     - (
-        ha_cluster_resource_defaults
+        ha_cluster_resource_defaults | length > 0
         and not 'pcmk.properties.resource-defaults.multiple'
         in __ha_cluster_pcs_capabilities
       ) or (
-        ha_cluster_resource_operation_defaults
+        ha_cluster_resource_operation_defaults | length > 0
         and not 'pcmk.properties.operation-defaults.multiple'
         in __ha_cluster_pcs_capabilities
       )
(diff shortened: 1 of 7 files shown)


richm:
@tomjelinek note that the old code actually mutated the internal data structure https://github.com/linux-system-roles/ha_cluster/pull/286/files#diff-c9af0ce31694e27416de4505d66e9823e982e151d5465d3c186047a3bbbeb76bR56 - it writes the `name` member of each certificate object in `ha_cluster_pcsd_certificates` - the new code does not - it only constructs the correct format for `certificate_requests` but does not touch `ha_cluster_pcsd_certificates` - I think this is ok - I can't see where `ha_cluster_pcsd_certificates` is used after this where the updated `name` value would be significant.

richm:
[citest]

PR #297, opened 2025-07-29
test: use post quantum crypto for certs where available

richm:
Test post quantum crypto by using certs generated by openssl where available.


richm:
[citest]

spetrosi (approved):

PR #304, opened 2025-08-13
test: clean up certs/keys created by tests

richm:
We are seeing test failures on some platforms caused by leftover keys and certs.
Make sure to clean these up, and make sure the test that uses the
certificate system role starts with no certs/keys.

In addition - on el9 systems we need to set the crypto policy to be able
to use Post Quantum Crypto (PQC) in tests.

## Summary by Sourcery

Clean up all test-generated certificates, keys, and authkeys before and after tests to eliminate leftover artifacts and prevent intermittent failures; add failure-rescue steps for log collection; and refine PSKS fixture key generation for distro compatibility

Bug Fixes:
- Remove leftover pcsd TLS certificates, keys, and PSKS authkeys in multiple test playbooks to fix intermittent failures

Enhancements:
- Introduce rescue tasks in tests to dump systemd, pcsd, and pacemaker logs on failure
- Load distribution variables in the PSKS fixture and default to rsa:2048 when mldsa65 is unsupported

Tests:
- Stop getcert tracking of the pcsd TLS certificate in cert-role tests to avoid stale tracking

### DIFF
--- a/.ostree/roles-testing.txt
+++ b/.ostree/roles-testing.txt
@@ -1,0 +1,1 @@
+crypto_policies
--- a/tests/collection-requirements.yml
+++ b/tests/collection-requirements.yml
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: MIT
+---
+collections:
+  - fedora.linux_system_roles
--- a/tests/tasks/fixture_psks.yml
+++ b/tests/tasks/fixture_psks.yml
@@ -29,6 +29,14 @@
       changed_when: false
       no_log: true  # this is quite verbose
 
+    - name: Load test variables
+      include_vars:
+        file: vars/rh_distros_vars.yml
+      when: __ha_cluster_is_rh_distro is not defined
+
+    # Seems like pcs/pcsd is not happy with mldsa65 on RHEL 9,
+    # even though it's supported by openssl.
+    # So we use rsa:2048 instead.
     - name: Generate a self signed pcsd cert and the pcsd key
       command: >-
         openssl req -x509 -newkey {{ key_algo }} -nodes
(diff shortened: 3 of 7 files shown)


richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

tomjelinek:
@richm You added two tasks at the end of `tests/tasks/fixture_psks.yml` to remember the current crypto-policy and change it to enable PQC. However, they seem to be switched - the policy is changed first and then the current (instead of original) policy is remembered. Therefore, the original policy is never restored.

On top of that, the policy is changed to enable PQC only if RHEL == 9, while 'Generate a self signed pcsd cert and the pcsd key' task has been modified to only generate PQC certificates for RHEL > 9.

Since these two changes contradict each other, I think one of them should be undone. I'm not sure what your intention was and which one should be kept, though.

Note that setting crypto-policy to DEFAULT:PQ fails for older RHEL 9 releases, effectively breaking our CI.

richm:
> @richm You added two tasks at the end of `tests/tasks/fixture_psks.yml` to remember the current crypto-policy and change it to enable PQC. However, they seem to be switched - the policy is changed first and then the current (instead of original) policy is remembered. Therefore, the original policy is never restored.
> 
> On top of that, the policy is changed to enable PQC only if RHEL == 9, while 'Generate a self signed pcsd cert and the pcsd key' task has been modified to only generate PQC certificates for RHEL > 9.
> 
> Since these two changes contradict each other, I think one of them should be undone. I'm not sure what your intention was and which one should be kept, though.
> 
> Note that setting crypto-policy to DEFAULT:PQ fails for older RHEL 9 releases, effectively breaking our CI.

https://github.com/linux-system-roles/ha_cluster/pull/306

PR #306, opened 2025-08-19
test: fix setting and resetting PQC

richm:
1. Only set PQC on RHEL 9.7 and later, but not RHEL 10 and later, and on EL9 platforms other than RHEL 9
2. Run the crypto_policies role with the variable set to null to get the current policy for resetting
later.

## Summary by Sourcery

Fix PQC test fixtures by targeting only supported RHEL/EL9 versions, retrieving and storing the existing crypto policy for cleanup, and standardizing the restore condition syntax

Bug Fixes:
- Restrict PQC enablement to RHEL 9.7 and later (but before RHEL 10) or non-RedHat EL9 variants to avoid unsupported platforms

Enhancements:
- Capture the current crypto policy by running the crypto_policies role with a null setting before enabling PQC
- Store the original crypto policy in a variable for post-test cleanup

Tests:
- Adjust the restore step’s when condition to consistently use double quotes for empty-string checks

### DIFF
--- a/tests/tasks/fixture_psks.yml
+++ b/tests/tasks/fixture_psks.yml
@@ -1,5 +1,49 @@
 # SPDX-License-Identifier: MIT
 ---
+- name: Load test variables
+  include_vars:
+    file: vars/rh_distros_vars.yml
+  when: __ha_cluster_is_rh_distro is not defined
+
+- name: Check if the managed node needs crypto-policies to be able to use PQC
+  when:
+    - __ha_cluster_is_rh_distro | bool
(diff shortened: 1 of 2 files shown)


richm:
[citest_bad]

tomjelinek:
The crypto-policy fix looks good.

However, it seems PQC is not used on RHEL 9 due to this condition in `tests/tasks/fixture_psks.yml`:
```
key_algo: "{{ 'mldsa65' if 'MLDSA65' in openssl_algorithms.stdout
  and __ha_cluster_is_rh_distro and ansible_facts['distribution_major_version'] | int > 9
  else 'rsa:2048' }}"
```

I'm running the patch through CI right now to confirm both statements.

richm:
[citest]

PR #376, opened 2026-04-24
feat: export stonith levels

idevat:
Enhancement:
Add stonith levels to ha_cluster_info module exporting current cluster configuration.

Reason:
This is the next step in implementing an info module which exports cluster configuration in a variables structure in the same format as ha_cluster role accepts.

Result:
Variables have been added to the export:
* `ha_cluster_stonith_levels`

https://redhat.atlassian.net/browse/RHEL-46232

## Summary by Sourcery

Export existing cluster stonith levels through the ha_cluster_info module and verify they round-trip with the role configuration.

New Features:
- Expose ha_cluster_stonith_levels from existing cluster configuration via ha_cluster_info when pcs supports stonith level JSON output.

Enhancements:
- Extend pcs capability detection and loader/exporter logic to handle stonith level configuration, including mapping pcs JSON output into role variable structure.

Documentation:
- Document stonith levels (fencing topology) as part of the configuration exported by ha_cluster_info.

Tests:
- Add integration and unit tests to validate stonith levels export, including various targeting modes and error handling for pcs failures.

### DIFF
--- a/library/ha_cluster_info.py
+++ b/library/ha_cluster_info.py
@@ -152,6 +152,7 @@ class Capability(Enum):
         "pcmk.properties.operation-defaults.config.output-formats"
     )
     CONSTRAINTS_OUTPUT = "pcmk.constraint.config.output-formats"
+    STONITH_LEVELS_OUTPUT = "pcmk.stonith.levels.config.output-formats"
 
 
 def get_cmd_runner(module: AnsibleModule) -> loader.CommandRunner:
@@ -411,6 +412,28 @@ def export_constraints_configuration(
     return result
 
 
+def export_stonith_levels_configuration(
+    module: AnsibleModule, pcs_capabilities: List[str]
+) -> Dict[str, Any]:
+    """
+    Export existing HA cluster stonith levels
+    """
+
+    if Capability.STONITH_LEVELS_OUTPUT.value not in pcs_capabilities:
+        return dict()
+
+    cmd_runner = get_cmd_runner(module)
+    stonith_levels = loader.get_stonith_levels_configuration(cmd_runner)
+
+    result: dict[str, Any] = dict()
+
+    levels = exporter.export_stonith_levels(stonith_levels)
+    if levels:
+        result["ha_cluster_stonith_levels"] = levels
+
+    return result
+
+
 def get_pcs_capabilities(module: AnsibleModule) -> List[str]:
     """
     Extract pcsd pcs_capabilities from pcs version info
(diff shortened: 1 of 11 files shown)


module_utils/ha_cluster_lsr/info/exporter.py
```diff
@@ -44,6 +44,9 @@
     export_resource_group_list,
     export_resource_primitive_list,
 )
+from .exporter_package.stonith_levels import (
+    export_stonith_levels,
+)
```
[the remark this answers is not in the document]
idevat: The intention of this module is to export selected functionality from exporter package (since `__init__.py` must be empty) not to using the imported variables.

tests/unit/test_ha_cluster_info_stonith_levels.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+
+    def test_no_capabilities(self) -> None:
+        with mocked_module([]) as module_mock:
+            self.assertEqual(
+                ha_cluster_info.export_stonith_levels_configuration(
+                    module_mock, pcs_capabilities=[]
+                ),
+                {},
+            )
```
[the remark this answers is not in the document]
idevat: This is over-engineering. The key you need is either on the list or it's not. It doesn't matter if it's Tuesday or Wednesday.

tomjelinek (approved):

library/ha_cluster_info.py
```diff
@@ -411,6 +415,28 @@ def export_constraints_configuration(
...
+    Export existing HA cluster stonith levels
+    """
+
+    if Capability.STONITH_LEVELS_OUTPUT.value not in pcs_capabilities:
+        return dict()
+
+    cmd_runner = get_cmd_runner(module)
+    stonith_levels = loader.get_stonith_levels_configuration(cmd_runner)
+
+    result: dict[str, Any] = dict()
```
tomjelinek, line 431:
```suggestion
    result: Dict[str, Any] = dict()
```

tomjelinek (changes requested):

library/ha_cluster_info.py
idevat: Why? What about:
* https://github.com/linux-system-roles/ha_cluster/blob/ca7a9e8446619f09c4969a09b27b1b6e77a0f000/library/ha_cluster_info.py#L176
* https://github.com/linux-system-roles/ha_cluster/blob/ca7a9e8446619f09c4969a09b27b1b6e77a0f000/library/ha_cluster_info.py#L227
* etc.
There are many similar issues.

idevat: If it is a real issue, we should create separate ticket, I guess.

tomjelinek: AI pointed out this is an issue in Python 3.6. You haven't responded to it, as you usually do, so I wanted to make sure this doesn't go unnoticed.

Since this is not the only occurrence, we can deal with it in a separate PR.

tomjelinek (approved):

richm (approved):

richm:
[citest]

PR #385, opened 2026-05-12
fix: check for corosync-qdevice service in bad state before stopping

richm:
Due to a weird quirk in the service_facts module, it can report that
the status is one of the bad_states instead of just leaving it out, when the service is
not installed. Therefore, we check for the status and if it is one of the bad_states, we
assume the service is not installed, and do not try to stop it.

## Summary by Sourcery

Bug Fixes:
- Avoid attempting to stop corosync-qdevice when service_facts reports it in a bad state that actually indicates it is not installed.


richm:
[citest]

tomjelinek (approved):

spetrosi (approved):

PR #388, opened 2026-05-18
refactor: use ansible.posix 2.1.X for EL7 compatibility [citest_skip]

richm:
The recently released ansible.posix 2.2.0 does not work on EL7.
Pin the version of ansible.posix to 2.1.X.

NOTE: Even though this role might not support EL7, this update
is applied to all system roles for consistency.  Plus, when this
role is part of the system roles collection, all roles must use
the same version of ansible.posix - there is no way for a role
which is part of a collection to use a different version of a
dependency than the version used by the other roles.

### DIFF
--- a/meta/collection-requirements.yml
+++ b/meta/collection-requirements.yml
@@ -2,6 +2,7 @@
 ---
 collections:
   - name: ansible.posix
+    version: '>=2.1.0,<2.2.0'
   - name: community.general
     version: '>=6.6.0,<12.0.0'
   - name: fedora.linux_system_roles


tomjelinek (commented):
This role does not work on EL7 anyway. Perhaps this is needed for purposes of collection / roles packaging, and must be present in all roles? A comment explaining the version pinning would be nice, though.

richm:
> This role does not work on EL7 anyway. Perhaps this is needed for purposes of collection / roles packaging, and must be present in all roles?

For the sake of consistency, yes, it just makes it easier if all roles use the same version.  Note that when the roles are part of a collection, we cannot use different versions of ansible.posix in different roles - they must all use the same version, which in this case must be 2.1.X.

> A comment explaining the version pinning would be nice, though.

Check out the latest comment.

PR #410, opened 2026-08-04
test: add ha_cluster_test_skip_kernel_modules to skip modules used by test machines

richm:
Some of our test systems use kernel modules that ha_cluster cannot unload.  Add
ha_cluster_test_skip_kernel_modules which is a string containing a space delimited
list of names of kernel modules to skip or ignore.  This will be passed in by the
test framework on those machines which require modules that cannot be unloaded.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBD watchdog module handling by excluding configured kernel modules from unloading.
  * Updated validation to accommodate intentionally skipped watchdog modules.

### DIFF
--- a/tasks/shell_pcs/sbd.yml
+++ b/tasks/shell_pcs/sbd.yml
@@ -27,11 +27,12 @@
               community.general.modprobe:
                 name: "{{ module }}"
                 state: absent
-              loop: "{{
-                __ha_cluster_local_node.sbd_watchdog_modules_blocklist | d([])
-                }}"
+              loop: "{{ __blocklist | difference(__test_skip_list) | list }}"
               loop_control:
                 loop_var: module
+              vars:
+                __blocklist: "{{ __ha_cluster_local_node.sbd_watchdog_modules_blocklist | d([]) }}"
+                __test_skip_list: "{{ (ha_cluster_test_skip_kernel_modules | d('')).split(' ') | list }}"
 
         - name: Configure and load watchdog kernel module
           block:
--- a/tests/template_sbd_all_options.yml
+++ b/tests/template_sbd_all_options.yml
@@ -87,6 +87,7 @@
       assert:
         that:
           - "'iTCO_wdt' not in __test_sbd_watchdog_sbd_lsmod.stdout"
+      when: not "iTCO_wdt" in ha_cluster_test_skip_kernel_modules | d("")
 
     - name: Check lsmod output for SBD watchdog module
       assert:


richm:
[citest]

richm:
[citest]

tomjelinek:
It looks to me as if production code is being modified for the sake of tests. I'm not a fan of such approach. I'd like to discuss an alternative:

Currently, `iTCO_wdt` is hardcoded in `tests/template_sbd_all_options.yml` to match the value of `sbd_watchdog_modules_blocklist` role variable set in SBD tests. If `iTCO_wdt` is extracted into a variable, test framework could set it to a different value - a name of a kernel module which can be unloaded. This variable would be used in `sbd_watchdog_modules_blocklist` and `tests/template_sbd_all_options.yml`. If there is no such module, then leave the variable empty and skip testing unloading kernel modules. The change proposed in this PR skips it anyway, so it wouldn't make much of a difference.

spetrosi:
> It looks to me as if production code is being modified for the sake of tests. I'm not a fan of such approach. 

I agree, sometimes it's simply not possible to avoid such approach due to ansible limitations, but if it is possible I'd like to avoid changing runtime code.

richm:
> It looks to me as if production code is being modified for the sake of tests. I'm not a fan of such approach. I'd like to discuss an alternative:
> 
> Currently, `iTCO_wdt` is hardcoded in `tests/template_sbd_all_options.yml` to match the value of `sbd_watchdog_modules_blocklist` role variable set in SBD tests. If `iTCO_wdt` is extracted into a variable, test framework could set it to a different value - a name of a kernel module which can be unloaded.

What exactly is the test doing a test for?  Is it testing that some kernel module can be unloaded?

> This variable would be used in `sbd_watchdog_modules_blocklist` and `tests/template_sbd_all_options.yml`. If there is no such module, then leave the variable empty and skip testing unloading kernel modules. The change proposed in this PR skips it anyway, so it wouldn't make much of a difference.

Ok.

tomjelinek:
> > It looks to me as if production code is being modified for the sake of tests. I'm not a fan of such approach. I'd like to discuss an alternative:
> > Currently, `iTCO_wdt` is hardcoded in `tests/template_sbd_all_options.yml` to match the value of `sbd_watchdog_modules_blocklist` role variable set in SBD tests. If `iTCO_wdt` is extracted into a variable, test framework could set it to a different value - a name of a kernel module which can be unloaded.
> 
> What exactly is the test doing a test for? Is it testing that some kernel module can be unloaded?

SBD uses hardware watchdogs. For the watchdogs to be available, their kernel modules must be loaded. Blocking certain watchdog modules also seems to be a use case. (This comes from a community contribution https://github.com/linux-system-roles/ha_cluster/commit/a5cbd7c7c55a383f3822547bbb2697d8ce55b156 so I'm not sure about the use case details.)

The test is doing a test for enabling a module (`softdog`) and disabling a module (`iTCO_wdt`). Both are two step operations: load / unload a module and create a config file so that the module is / is not loaded on boot.

Actually, the test is doing a test for the complete SBD configuration, but there is no issue with that. And changing the test so that it does not unload a module should not affect the rest of the test.

richm:
closing in favor of https://github.com/linux-system-roles/ha_cluster/pull/411

PR #412, opened 2026-08-06
feat: Write roles fingerprints to /var/log/sysroles.jsonl [citest_skip]

spetrosi:
Feature: Write roles fingerprints to /var/log/sysroles.jsonl [citest_skip]

Reason: By default logs are printed to rsyslog. This change adds a possibility to write logs to a file on the system for the downstream users.

Result: For the upstream, this makes rsyslog log message more detailed. For the downstream - also writes logs to /var/log/sysroles.jsonl

### DIFF
--- a/pytest_extra_requirements.txt
+++ b/pytest_extra_requirements.txt
@@ -1,4 +1,8 @@
 # SPDX-License-Identifier: MIT
 
-# ansible and dependencies for all supported platforms
-ansible-core ; python_version > "2.6"
+# Write extra requirements for running pytest here:
+# If you need ansible then uncomment the following line:
+-ransible_pytest_extra_requirements.txt
+# If you need mock then uncomment the following line:
+#mock ; python_version < "3.0"
+dacite
--- a/pytest_extra_requirements.txt
+++ b/pytest_extra_requirements.txt
@@ -1,8 +1,4 @@
 # SPDX-License-Identifier: MIT
 
-# Write extra requirements for running pytest here:
-# If you need ansible then uncomment the following line:
--ransible_pytest_extra_requirements.txt
-# If you need mock then uncomment the following line:
-#mock ; python_version < "3.0"
-dacite
+# ansible and dependencies for all supported platforms
+ansible-core ; python_version > "2.6"
--- a/library/sr_fingerprint.py
+++ b/library/sr_fingerprint.py
@@ -7,29 +7,148 @@ __metaclass__ = type
 DOCUMENTATION = """
 ---
 module: sr_fingerprint
-short_description: Write a message string to syslog using Ansible C(module.log) function.
+short_description: Write role fingerprint data to syslog and optionally to a JSONL log file
 description:
-    - Writes the given string to the system log using Ansible C(module.log) function.
+    - Collects role fingerprint data into a canonical record and writes it to
+      syslog using Ansible C(module.log) as C(key=value) pairs.
+    - Optionally appends the same record as a JSON line to a log file
+      (one JSON object per line, JSONL format), by default
+      C(/var/log/sysroles.jsonl).
+    - Playbook variables are not available inside modules automatically. Roles
+      pass C(role_name), C(role_path), C(ansible_play_hosts_all),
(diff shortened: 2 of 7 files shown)


tomjelinek:
mypy and pylint report a lot of errors in `tests/unit/test_sr_fingerprint.py`. If you do not intend to fix them, add the file to ignores in `mypy.ini` and `pylintrc`, please.

black reports the file is not formatted properly, please run `black --config pyproject.toml tests/unit/test_sr_fingerprint.py` to fix that.

richm:
> mypy and pylint report a lot of errors in `tests/unit/test_sr_fingerprint.py`. If you do not intend to fix them, add the file to ignores in `mypy.ini` and `pylintrc`, please.

Hmm - each role having its own python coding standards is a recipe for chaos, especially when we try to roll out code common to all roles.  I guess we will need to figure out which roles have the strictest standards and enforce those across all roles.  And ensure that we have proper github action checks for all of these.

I don't see these pylint and mypy errors.  How can we add a github action workflow for them?

> 
> black reports the file is not formatted properly, please run `black --config pyproject.toml tests/unit/test_sr_fingerprint.py` to fix that.

Is that only because of the `line_length = 80` or something else?

richm:
system roles generally uses tox-lsr https://github.com/linux-system-roles/tox-lsr/tree/main/src/tox_lsr/config_files to enforce our coding standards

Now - maybe the right answer is "hey, the lsr coding standards are too old and/or too lax and need to be updated" - in that case, we may need to schedule several days worth of work to update the python code used in most of our 34 system roles

tomjelinek:
> > mypy and pylint report a lot of errors in `tests/unit/test_sr_fingerprint.py`. If you do not intend to fix them, add the file to ignores in `mypy.ini` and `pylintrc`, please.
> 
> Hmm - each role having its own python coding standards is a recipe for chaos, especially when we try to roll out code common to all roles.

Agreed. I'm fine with adding such common files to an ignore list so the linters don't report them.

> I guess we will need to figure out which roles have the strictest standards and enforce those across all roles. And ensure that we have proper github action checks for all of these.
> 
> I don't see these pylint and mypy errors. How can we add a github action workflow for them?

Those are defined in `tmt/tests/linters.fmf` and that is run by `.gitlab-ci.yml`, so it is not executed in the PR here. Originally, there was no common python code, so we added these tasks to run in our private CI to cover python code specific to this role.

> 
> > black reports the file is not formatted properly, please run `black --config pyproject.toml tests/unit/test_sr_fingerprint.py` to fix that.
> 
> Is that only because of the `line_length = 80` or something else?

It looks like it's just the line length. And there is actually an ignore list for black in `pyproject.toml`.

I'm fine with adding the files to existing ignore lists.

spetrosi:
[citest]

pytest_extra_requirements.txt
```diff
@@ -1,8 +1,4 @@
 # SPDX-License-Identifier: MIT
 
-# Write extra requirements for running pytest here:
-# If you need ansible then uncomment the following line:
--ransible_pytest_extra_requirements.txt
-# If you need mock then uncomment the following line:
-#mock ; python_version < "3.0"
-dacite
+# ansible and dependencies for all supported platforms
+ansible-core ; python_version > "2.6"
```
richm: @spetrosi ugh - this is one of those that we cannot just replace with a standard one from .github :-(

In this case, the file `ansible_pytest_extra_requirements.txt` provides the necessary `ansible-core` requirement.

richm: This should have been flagged by running the `.github` automation with `lsr_dry_run=true`

spetrosi: I tried running with lsr_dry_run but the output was so large that I couldn't read it. I'll check each role for the changes in pytest_extra_requirements.txt now.

spetrosi: I will revert this in roles where this change was breaking, and will remove this file from .github so that it is managed by each role

spetrosi:
[citest]

richm (approved):

